PR #2, opened 2015-03-11
rust: replace pushd usage in rust do_install

posborne:
In dizzy, the following error was encountered when running do_install
on the rust recipe:

    pushd: not found

According to the bitbake user manual, /bin/sh is used and bashisms
(like pushd) should be avoided.  In my case, /bin/sh turns out to
be zsh (with no extra configuration, so pushd is not present).

Signed-off-by: Paul Osborne <osbpau@gmail.com>

### DIFF
--- a/recipes/rust/rust.inc
+++ b/recipes/rust/rust.inc
@@ -358,9 +358,7 @@ rust_do_install () {
 
         ## rust will complain about multiple providers of the runtime libs
         ## (libstd, libsync, etc.) without this.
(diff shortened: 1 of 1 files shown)


posborne:
Here's the relevant portion of the user manual: http://www.yoctoproject.org/docs/1.6/bitbake-user-manual/bitbake-user-manual.html#shell-functions

jmesmon:
Thanks for both of the PRs.

PR #5, opened 2015-04-29
Upgrade to rust beta

srwalter:
Upgrade rust to 1.0.0-beta, and update cargo so that it will build with the -beta compiler.

Add snapshots to SRC_URI to avoid network traffic during do_compile

Break rust-cross' dependency on rust-native.  This reduces by half the amount of time needed to build the cross compiler.


classes/cargo.bbclass
```diff
@@ -89,6 +89,7 @@ oe_cargo_fix_env () {
 	export HOST_CC="${RUST_BUILD_CC}"
 	export HOST_CFLAGS="${RUST_BUILD_CFLAGS}"
 	export HOST_AR="${BUILD_AR}"
+	export RUST_SYSROOT="${STAGING_DIR_NATIVE}/${prefix_native}"
```
jmesmon: Is this sysroot (pointing to native) always appropriate? Even if we're building for TARGET?

Also, it looks like this env var is used by rustc itself (the reset of these are specific to gcc-rs). Is there a reason it isn't simply exported in the rust.bbclass?

classes/rust.bbclass
```diff
@@ -1,7 +1,6 @@
 RUSTC = "rustc"
 
-# FIXME: --sysroot might be needed
-RUSTC_ARCHFLAGS += "--target=${TARGET_SYS} -C rpath"
+RUSTC_ARCHFLAGS += "--target=${TARGET_SYS} --sysroot=${STAGING_DIR_NATIVE}/${prefix_native} -C rpath"
```
jmesmon: Again, is this sysroot always correct?

What type of failure did not having it trigger?

recipes/cargo/cargo.inc
```diff
@@ -14,8 +14,12 @@ DEPENDS = "openssl zlib libgit2 curl ca-certificates libssh2"
 SRC_URI = "\
 	git://github.com/rust-lang/cargo.git;protocol=https;name=cargo \
 	git://github.com/rust-lang/rust-installer.git;protocol=https;name=rust-installer;destsuffix=git/src/rust-installer \
+	http://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-04-02/cargo-nightly-x86_64-unknown-linux-gnu.tar.gz \
```
jmesmon: Having this as an option makes sense, but I'd really rather it not be mandatory.

Could you add a PACKAGECONFIG setting that allows us to chose between using the rustc-snapshot & depending on a local rustc build?

jmesmon:
I'm not sure I understand why the patches for libdir & bindir support aren't used any more.

I was fairly certain bitbake puts -cross packages in the same prefix as native ones but with different bindir's and libdirs, which means rust-native package will conflict with rust-cross.

I suppose it's possible you didn't run into this due to using snapshots for everything.

classes/cargo.bbclass
srwalter: Yes, we can always use this value as the sysroot.  Rust is smart enough to look in a ${RUST_SYSROOT}/${TARGET_SYS} when cross-compiling.

This could be moved to rust.bbclass, but it's not necessary.  That bbclass already throws the --sysroot command flag, which I find preferable due to requiring less action-at-a-distance.  Unfortunately that is required we cargo, as we have no way to get cargo to throw the --sysroot flag.

classes/rust.bbclass
srwalter: Without this, I was seeing failures along the lines of "failed to find target x86_64-linux.json".  The --bindir patch was previously causing the default sysroot to be correct, however I had difficulties with that patch, see below...

recipes/cargo/cargo.inc
srwalter: There's already the code in cargo.inc for "rust-snapshot", though I think that's a copy/paste error since it should be for a cargo-snapshot, right?

srwalter:
The patches I removed were for some combination of not applying, being unnecessary, and being actively harmful.  I think you'd agree that it's ideal to patch rust as minimally as possible for ease of maintenance.  The bindir patch in particlar was somewhat invasive, just due to having to plumb that value all through the build system.

I did test building rust-native and rust-cross in the same workspace.  Both worked, and I got no errors.  I would have expected bitbake to warn/error out if both recipes tried to install the same file to the sysroot.  In general that is bitbake's behavior.

Note that I actually maintained the same installed behavior of putting rust-cross in ${bindir}/${TARGET_SYS}, see the lines I added to do_install.  It's simply much less invasive to let rust install however it wants, then move the files around to our desired layout.

recipes/cargo/cargo.inc
jmesmon: Ah, I missed that this was for cargo. Nevermind, that seems ok (well, it only works for x86-64, but we can tweak that later).

recipes/rust/rust.inc
```diff
@@ -337,6 +328,8 @@ rust_runmake () {
 }
 
 do_compile () {
+	mkdir -p dl
+	cp -f ${WORKDIR}/rust-stage0-2015-03-27-5520801-linux-x86_64-ef2154372e97a3cb687897d027fd51c8f2c5f349.tar.bz2 dl
```
jmesmon: This path looks ideal for a variable.

recipes/rust/rust.inc
```diff
@@ -345,6 +338,8 @@ rust_do_install () {
 
         # Rust's install.sh doesn't mark executables as executable because
         # we're using a custom bindir, do it ourselves.
+	mkdir -p ${D}/${bindir}/
+	find ${D}/${prefix}/bin -type f -exec mv \{} ${D}/${bindir} \;
         chmod +x "${D}/${bindir}/rustc"
         chmod +x "${D}/${bindir}/rustdoc"
         chmod +x "${D}/${bindir}/rust-gdb"
```
jmesmon: Since bindir isn't being used anymore, I'm pretty sure we can let rust's installer handle chmoding these

recipes/rust/rust.inc
srwalter: Agreed, I'll fix that and update the PR

srwalter:
Fixes pushed

jmesmon:
I went through this today and cherry-picked a few commits, modified and commited a few, and dropped a couple.

Notably, I've stuck with my changes to teach rustc about libdirs and bindirs rather than using RUST_SYSROOT. While I do think it would be a good idea to keep the patches as simple as possible, I was running into problems with the sysroot related changes where rust would link itself agaist the system rustlib instead of the local one being built, leading to unexpected failures. Additionally, while the changes to support bindir and libdir properly are more invasive, I'm more comfortable dealing with them at this point and fairly confident they correctly handling things.

Thanks for this PR, please take a look at the latest master and see if it functions as expected for you, and let me know if I've inadvertantly not pulled something in that I should have.

PR #8, opened 2015-07-01
Make LLVM in its own recipe

srwalter:
rust takes a painfully long time to build.  A non-trivial part of that
is building LLVM.  Fortuitiously, the LLVM component is common between
the -native compiler and all variants of the -cross compiler.  That
means we can build it once, and reuse the output for any flavor of rust.
That saves build time, and it utilizes bitbake's sstate more
efficiently.  It's win/win!

### DIFF
--- a/recipes/rust/rust-llvm-native_1.1.0.bb
+++ b/recipes/rust/rust-llvm-native_1.1.0.bb
@@ -1,0 +1,27 @@
+SUMMARY = "LLVM compiler framework (packaged with rust)"
+LICENSE = "NCSA"
+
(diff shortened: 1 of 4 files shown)


srwalter:
Any chance I could get some comment on this?  I actually have several more merge requests to make, I just didn't want to bombard you with them all at once :-)

jmesmon:
@srwalter I'm testing it right now, it looks good. Should be merged shortly

jmesmon:
@srwalter thanks for this. I've build just finished up and looks good. I've added a commit that enables debug symbols in llvm to avoid some complaints about pre-stripping in poky.

PR #9, opened 2015-07-06
Disable symbol hashing

srwalter:
Rust's symbol hashes break sstate, because a subsequent build of the same
input hash may generate different symbol hashes.  This means that
packages that previously built against the "same" (from bitbake's
perspective) rustc would no longer run, and bitbake would not know to
rebuild them because the input hash is the same.

Fortunately, bitbake's input hash makes rust's symbol versions
unnecessary; any change to rust's source or configuration will cause
rustc and all dependencies to rebuild.  Therefore, it's safe to defeat
rust's symbol versions entirely with a static version.

### DIFF
--- a/recipes/rust/files/rust-1.1.0/0009-src-rt-arch-i386-morestack.S-call-rust_stack_exhaust.patch
+++ b/recipes/rust/files/rust-1.1.0/0009-src-rt-arch-i386-morestack.S-call-rust_stack_exhaust.patch
@@ -1,0 +1,47 @@
+From 02936deefb786a244c98ec2293e6b85ecfc85cdb Mon Sep 17 00:00:00 2001
+From: Steven Walter <swalter@lexmark.com>
+Date: Mon, 15 Jun 2015 11:35:47 -0400
+Subject: [PATCH] src/rt/arch/i386/morestack.S: call rust_stack_exhausted via
+ plt
+
+This prevents a relocation in the text section.  Text relocations are
+incompatible with hardened kernels.
+
+https://github.com/rust-lang/rust/issues/5714
+---
+ src/rt/arch/i386/morestack.S | 5 ++---
+ 1 file changed, 2 insertions(+), 3 deletions(-)
+
+diff --git a/src/rt/arch/i386/morestack.S b/src/rt/arch/i386/morestack.S
+index 2f03251..98fdfdf 100644
+--- a/src/rt/arch/i386/morestack.S
++++ b/src/rt/arch/i386/morestack.S
+@@ -75,7 +75,7 @@
+ #else
+ #if defined(__linux__) || defined(__FreeBSD__)
+ #define MORESTACK               __morestack
+-#define EXHAUSTED               rust_stack_exhausted
++#define EXHAUSTED               rust_stack_exhausted@plt
+ #else
+ #define MORESTACK               ___morestack
+ #define EXHAUSTED               _rust_stack_exhausted
+@@ -83,7 +83,6 @@
+ #endif
+ 
+ .globl MORESTACK
+-.globl EXHAUSTED
+ 
+ // FIXME: What about __WIN32__?
+ #if defined(__linux__) || defined(__FreeBSD__)
+@@ -138,7 +137,7 @@ MORESTACK:
+ 
+         // re-align the stack
+         subl $12,%esp
+-        calll EXHAUSTED
++        call EXHAUSTED
+         // the exhaustion function guarantees that it can't return
+ 
+ 	.cfi_endproc
+-- 
+1.9.1
+
--- a/recipes/rust/files/rust-1.1.0/0010-disable-symbol-version-hashing.patch
+++ b/recipes/rust/files/rust-1.1.0/0010-disable-symbol-version-hashing.patch
@@ -1,0 +1,26 @@
+From 9db9d253355cc0e9e885bb1e63b054e318f9395a Mon Sep 17 00:00:00 2001
+From: Steven Walter <swalter@lexmark.com>
+Date: Tue, 7 Jul 2015 14:57:42 -0400
+Subject: [PATCH] Remove crate metadata from symbol hashing
+
+---
+ src/librustc_trans/back/link.rs | 3 ---
+ 1 file changed, 3 deletions(-)
+
+diff --git a/src/librustc_trans/back/link.rs b/src/librustc_trans/back/link.rs
+index 2f2ba16..ea3e897 100644
+--- a/src/librustc_trans/back/link.rs
++++ b/src/librustc_trans/back/link.rs
+@@ -205,9 +205,4 @@ fn symbol_hash<'tcx>(tcx: &ty::ctxt<'tcx>,
+     symbol_hasher.input_str(&link_meta.crate_name);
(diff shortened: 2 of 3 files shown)


jmesmon:
I'm not sure this is the right thing to do.

Reading the comment earlier in link.rs seems to say that the hash here is used for making symbols unique based on their type & crate:
```
 *  - Define STH(sym) as SHA256(CMH, type_str(sym))
```

If we make this hash a fixed value, what stops:

 - symbols with the same name but different types from colliding?
 - symbols with the same name in different crates (where the crates also have the same name but differ in their crate_id (also called a "package id" by the docs)?

jmesmon:
I'm also not quite sure how having this hash could break sstate: the comment implies that the hash is stable, so it shouldn't end up changing without source changes. In which case, I wouldn't expect bitbake to be unhappy as it is going to need to rebuild anyhow.

srwalter:
I believe the hash is stable for the same inputs, but the problem is that the inputs to the hash are not stable.  I believe the specific problem was the crate metadata (there's data in there that changes from run-to-run).  I'll do an experiment to see if it sufficient to just exclude crate metadata from the hash calculation.

srwalter:
Unfortunately, just removing the crate metadata does not produce a stable hash.  The problem is the crate id.  The crate id is randomly chosen whenever the crate is created, which causes the problem.

Would it be acceptable if I modified the hash algorithm to only include the type information and crate name (no crate metadata, no crate id)?

jmesmon:
I'm not sure I'm quite comfortable with that. I don't think I have a complete grasp on what's going on in the symbol hashing, but:

 - Other than comments, I couldn't find anything talking about the crate_id. It appears that attribute was changed to crate_name. When you say `crate_id` are you talking about the name? Or something else? I suppose another part of this is that it isn't clear to me where that randomness is being added, and if it is what it's purpose would be?

 - I'm concerned that eliminating the inclusion of these things will break some crates depending on slightly different variants of other crates (that may have the same version & name). I'm presuming the metadata includes enough info to differentiate them (I'm assuming the metadata knows where the crate is "from")

 - I suppose I don't really have "a fire lit under me" on this issue as I haven't seen it myself. Is there some way I can reproduce it locally? Or does it only show up occasionally? (Don't read this as me not being interested in getting this fixed, I'm very interested in this because I _really_ don't like getting build failures. I just don't want to leave a booby-trap for myself in the future by not making sure we're doing this the right way)

srwalter:
This is the updated patch I'm proposing:

http://pastebin.com/xxt0RwuT

So when I mentioned "crate_id" I should have said "crate_hash".  I haven't taken the time to track down exactly where the randomness originates, but it wouldn't surprise me if it's similar to what you said: a way to make crates with the same name differ in symbols.

I can't speculate as to exactly what may break from removing these values from the hash.  My general impression was that the purpose of the hash was to ensure that downstream packages got rebuilt after a compiler change (due to rust's lack of a stable ABI), but that purpose is served well by bitbake's input hashing.

You can pretty easily reproduce the hashes changes.  Use objdump --demangle -t to list the symbols on one of the libraries; I usually look for UdpSocket in libstd:

sys_common::net::UdpSocket::set_membership::h5924ccade134174dUOq

That's the full symbol name including hash in my local build.  If you rebuild the compiler (bitbake rust-native -C compile) and check the symbol again, you'll see that the hash at the end of the symbol name is different.  If you want to see the specific badness that this causes downstream, it requires a slightly more involved setup.  You need at least two crates, where A depends on B.

1) Build crates A and B
2) bitbake -c cleansstate rust-cross-i586 A # get rid of your sstate for the compiler and the end package
3) bitbake A # this will cause the compiler and A to rebuild, but not B

When A rebuilds, it will complain that B has references to symbols that don't exist.  That's because B did not relink against the new libstd, and so it's still looking for the symbol names from the old library.  If the symbol names in libstd were stable, this wouldn't have been a problem.

Hope that explanation helps!

posborne:
I'm curious if we end up having the same problems now that https://github.com/rust-lang/rfcs/blob/master/text/0109-remove-crate-id.md has been implemented with https://github.com/rust-lang/rust/pull/15319.

If the `crate_id` is in fact the source of entropy here, I think that change should do it.

srwalter:
Which "that change" are you referring to?  PR#15319 was already merged into 1.1.0, and I was still seeing unstable symbol hashes on 1.1.0.  Are you referring to my second patch on pastebin?

posborne:
Sorry for the confusion.  I was referring to PR#15319 and wasn't aware it was in 1.1.0.  Clearly, it doesn't fix the problem.

According to this (https://doc.rust-lang.org/rustc_back/svh/), the `crate_hash` (which is an `Svh`) should just be a hash on the AST.  I think we would have to try to capture the two ASTs in order to determine what is changing.  I *believe* your patch would be safe within the context of bitbake.

Looking at the original PR where Svh was added, you can see what a few sources of indeterminism were in the past: https://github.com/rust-lang/rust/pull/12533.  Hopefully, we can also figure that out and get a fix in upstream, so we don't need to maintain a patch like this long term.

srwalter:
I've updated this branch to only remove crate_hash and crate metadata from the hashing, as discussed.

I'll see what I can find in the AST that changes from run-to-run

jmesmon:
Primary concern with this is that I every now and then need to do various overriding (including crates via path in Cargo.toml) and (with this PR applied) I'm not sure what will happen if that causes 2 versions (which may have the same version number) of the same crate to be pulled in.

 - If it gives a link failure, I can probably be okay with this (but the first time I run into that failure I'm going to be fairly unhappy :) )
 - If it doesn't fail at link but instead arbitrarily chooses one of the symbols (essentially, a compilation), I really can't merge this.

So if we're sure the first is the case, I'm fine with merging for now. But if we're not sure, or we know the second thing will occur the PR will need to be tweaked.

srwalter:
An error I have seen during linking is "multiple candidates for libfoo".
That is, if rustc finds two libraries matching the name of the crate
requested, the link will fail.

On Sun, Jul 12, 2015 at 11:03 PM, Cody Schafer <notifications@github.com>
wrote:

> Primary concern with this is that I every now and then need to do various
> overriding (including crates via path in Cargo.toml) and (with this PR
> applied) I'm not sure what will happen if that causes 2 versions (which may
> have the same version number) of the same crate to be pulled in.
>
>    - If it gives a link failure, I can probably be okay with this (but
>    the first time I run into that failure I'm going to be fairly unhappy :) )
>    - If it doesn't fail at link but instead arbitrarily chooses one of
>    the symbols (essentially, a compilation), I really can't merge this.
>
> So if we're sure the first is the case, I'm fine with merging for now. But
> if we're not sure, or we know the second thing will occur the PR will need
> to be tweaked.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jmesmon/meta-rust/pull/9#issuecomment-120793096>.
>



-- 
-Steven Walter <stevenrwalter@gmail.com>

jmesmon:
Ok, I've merged this for now. I'm still concerned about preventing multiple versions of crates with the same name, but it doesn't seem like we're at risk of miss-comiling (only failing to build certain rust projects when we shouldn't fail).

PR #10, opened 2015-07-17
Add rustlib.bb and rust-bin.bbclass

srwalter:
rustlib.bb packages up the rust runtime for inclusion in the final image.  rust-bin.bbclass can be used to build rust libraries and binaries without cargo (because cargo can't play nice with bitbake's fetcher yet)

### DIFF
--- a/classes/rust-bin.bbclass
+++ b/classes/rust-bin.bbclass
@@ -81,13 +81,13 @@ oe_compile_rust_bin[vardeps] += "get_overlap_externs"
 oe_install_rust_lib () {
     for lib in $(ls ${LIBNAME}.{so,rlib} 2>/dev/null); do
         echo Installing $lib
-        install -D -m 644 $lib ${D}/${rustlibdir}/$lib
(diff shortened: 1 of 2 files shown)


jmesmon:
So just to clarify: the need for this is to allow things that link dynamically with the std lib? And the rust-bin bbclass is to allow building those things?

In looking at yocto, I've noticed that a similar item for gcc is named `gcc-runtime` (at least, I think it's similar), do we still want to prefer the name `rustlib` over something like `rust-runtime`?

srwalter:
Correct.  I'd be amenable to renaming it rust-runtime; I chose rustlib simply because that's the name of the directory where the standard libraries end up.  Let me know if that's what you'd prefer

jmesmon:
I suppose my thought was that other things might end up getting shipped in the rustlib dir, making the name of the package somewhat less exact in the future.

I'm fine with whatever way you want to go.

srwalter:
I'd propose we go ahead and merge this as-is, and keep the rename option on the table if it makes sense in the future.

jmesmon:
Ok, thanks.

PR #11, opened 2015-07-23
Add recipes for various rust libraries

srwalter:


jmesmon:
Can you elaborate on your plan with regards to adding crates as bitbake packages? 

I presume the reasoning is that it allows us to properly take advantage of bitbake's source caching mechanisms?

Is the intent to limit the crates to certain ones, or eventually have all the crates people are interested in using as bitbake packages?

srwalter:
That's correct, building this way allows bitbake to do all the fetching.  Cargo has an additional disadvantage inside of bitbake, which is that it wants to rebuild all dependencies on each run.  Suppose you have 3 crates, A depends on B depends on C.  When cargo builds B, build both B and C.  When cargo builds A, it will build, A, B, and C.  So, if you have a dependency tree of more than a couple levels, there is a significant amount of wasteful rebuilding occurring.

I haven't thought much about how or if to limit which crates.  At least until it becomes a problem, I would probably advocate accepting as many recipes as contributors are willing to provide :-)

PR #29, opened 2016-02-01
Add more crates

srwalter:
Add crate recipes for rustc-serialize, debug-builders, unix-socket, rand, and crypto

Additionally, fix up a couple of minor recipe bugs


jmesmon:
Thanks.

PR #32, opened 2016-02-02
Fix prelink issue breaking .bss

srwalter:
Rust programs included in prelinked images often crash at boot.

The reason prelink seemed to break relocations in .bss is because the
.note.rustc section happened to be placed in front of .bss. Using
objcopy to strip the note renumbered the sections and the symbols in the
.dynsym section were pointing to the wrong section index, but only for
.bss. Prelink would skip updating these dyn entries because it thought
they were pointing to a section that didn't get relocated, like
.comment. In some cases the GOT was actually prelinked correctly, but
the dynamic linker at run time would modify it to point at the
non-prelinked address.

Removing the SHF_ALLOC flag from the section causes it to be placed at
the end of the executable where it can be safely removed.

This is still a hack to the compiler because I couldn't find any way to
get the section attributes to be correct using the high-level LLVM
interface.

### DIFF
--- a/recipes/rust/files/rust-1.5.0/0010-rustc_trans-make-.note.rustc-look-more-like-debug-in.patch
+++ b/recipes/rust/files/rust-1.5.0/0010-rustc_trans-make-.note.rustc-look-more-like-debug-in.patch
@@ -1,0 +1,158 @@
+From 1bd3ab55ce24b2a54e021ba471a5f934b3b9ad6b Mon Sep 17 00:00:00 2001
+From: Steven Walter <swalter@lexmark.com>
+Date: Tue, 7 Jul 2015 16:49:44 -0400
+Subject: [PATCH 10/12] rustc_trans: make .note.rustc look more like debug info
+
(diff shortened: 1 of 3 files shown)


jmesmon:
> Using
objcopy to strip the note renumbered the sections and the symbols in the
.dynsym section were pointing to the wrong section index, but only for
.bss.

That sounds a lot like an objcopy bug. Or does it purposefully not fix section references when it removes a section?

srwalter:
Correct, fixing such references is beyond the scope of objcopy.  There could be references embedded within executable code, which objcopy would know nothing about.  That's why "debug" sections generally go at the end of the file, so that they can be removed with impunity.  Really, this patch is just about making the .note.rustc section act like other "optional" sections.

jmesmon:
Got it, thanks (though technically, the references aren't in code (.text), they are in sections that refer to .text, like the relocations).

For future reference, it looks like the upstream rustc issue is: https://github.com/rust-lang/rust/issues/26764

PR #33, opened 2016-02-03
rust: force crate hash to use bitbake input hash

srwalter:
The current crate hash is not stable from run-to-run.  This causes
problems with bitbake; it needs a guarantee that every build with the
same input will generate compatible output, otherwise sstate won't work.
Using -C crate_hash, we can do that by using the bitbake input hash to
determine the crate hash; the bitbake input hash will be stable, but
still different for different rust recipes.

This should eliminate the problems that have been seen when bitbake
rebuilds part of the rust stack (due to missing sstate) and ends up
getting hash mismatches.  Unlike rust's internally generated hash, ours
is guaranteed to be stable.

### DIFF
--- a/classes/rust.bbclass
+++ b/classes/rust.bbclass
@@ -1,7 +1,7 @@
 RUSTC = "rustc"
 
 # FIXME: --sysroot might be needed
-RUSTC_ARCHFLAGS += "--target=${TARGET_SYS} -C rpath"
+RUSTC_ARCHFLAGS += "--target=${TARGET_SYS} -C rpath -C crate_hash=${BB_TASKHASH}"
 
 def rust_base_dep(d):
     # Taken from meta/classes/base.bbclass `base_dep_prepend` and modified to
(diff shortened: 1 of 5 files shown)


srwalter:
(This'll be the last one for a bit :smile_cat: )

jmesmon:
Just to confirm, after adding this:

 - all the rustc provided crates will have the same crate_hash
 - this is OK because they all have different names

Is there an rustc ticket about the lack of reproducible builds that notes this issue?

jmesmon:
And how does this interact with the existing patch to disable including the crate_hash (and crate_metadata) in the symbol hashing? https://github.com/jmesmon/meta-rust/blob/master/recipes/rust/files/rust-1.5.0/0009-Remove-crate-metadata-from-symbol-hashing.patch

Does this change allow us to re-add the crate_hash there?

srwalter:
Correct, the existing patch for symbol hashing is still required.  That changes the name mangling on the symbols themselves, which is also important.  This change primarily impacts the name of the shared object itself.

Correct, it's okay for the rustc provided crates to have the same hash because they have different names.  The crate hash is to differentiate versions of same-named crates.

I'm not aware of an upstream ticket on this issue.

jmesmon:
Ok. It'd be nice to fix the reproducible build issues more permanently (without disabling and overriding things), but I'm OK with this for now.

PR #35, opened 2016-02-14
Fix sstate for multiple arch deployments

### LINKED ISSUE #15: With latest poky, rust cross builds no longer are suffixed with `${TARGET_ARCH}`
ie: `rust-cross-arm` becomes `rust-cross`.

This is due to some changes in how `BBCLASSEXTEND` and variable tracking work. Essentially, our `PN_class-cross` is being overridden by the internal `BBCLASSEXTEND` machinery.

Assuming the `${TARGET_ARCH}` suffix is required (it was just copied from how gcc's `-cross` does it), without a fix for this in `bitbake`, we'll need to switch to using seperate .bb files instead of `BBCLASSEXTEND`.

derekstraka:
Currently, the rust-cross package was not created for each target architecture which causes issues when you switch between MACHINE and have a shared SSTATE cache.  The set of changes includes:

* Adding the LLVM and new rust-cross pacakge to the set that disable PIE
* Adding an LLVM patch that fixes an issue when debugedit is executed.  The second '/' has already been removed upstream, but was not fixed in rust 1.5
* Create a new rust-cross-${TARGET_ARCH} package that is similar to how other cross compilers are built.  The main issue is that PN_class-cross is not honored in [ast.py](https://github.com/openembedded/bitbake/blob/master/lib/bb/parse/ast.py#L451), so the resulting package is not unique per target
* The debug info is removed from llvm since it was over 2GB of data and caused the build time to increase by over 5 minutes (resulting from copy + compress)

Fixes #15


conf/distro/include/rust_security_flags.inc
```diff
@@ -1,4 +1,6 @@
 # Build errors with PIE options enabled
 SECURITY_CFLAGS_pn-rust-native = "${SECURITY_NO_PIE_CFLAGS}"
+SECURITY_CFLAGS_pn-rust-cross-${TARGET_ARCH} = "${SECURITY_NO_PIE_CFLAGS}"
 SECURITY_CFLAGS_pn-rust-cross = "${SECURITY_NO_PIE_CFLAGS}"
```
jmesmon: Since we now have `rust-cross-${TARGET_ARCH}` should this line about `rust-cross` be removed?

jmesmon:
Just to confirm: testing this could be done by building `rust-hello-world` once, then changing the machine to one with a different arch and then attempting to build again?

derekstraka:
That's correct.  I validated with a x86_64 and aarch64 based build.

conf/distro/include/rust_security_flags.inc
derekstraka: That's true; I will remove that.  I'll wait to see if there are any more changes.

jmesmon:
Thanks, I've merged this along with an additional commit to remove the `rust-cross` reference.

PR #39, opened 2016-02-18
Allow the fetch to complete before attempting to patch

### LINKED ISSUE #38: Building rust-native from scratch causes errors
#15 introduced a patch that now causes issues during the build:

> WARNING: rust-llvm-native-1.5.0-r0 do_unpack: rust-llvm-native: the directory ${TMPDIR}/work-shared/${SOURCE_NAME}-${PV}-${PR}/rustc-${PV}/src/llvm (/var/build/build/tmp/work-shared/rust-1.5.0-r0/rustc-1.5.0/src/llvm) pointed to by the S variable doesn't exist - please set S within the recipe to point to where the source has been unpacked to

Update the shared-source-use.bbclass to block at patch rather than configure

derekstraka:
fixes #38

### DIFF
--- a/classes/shared-source-use.bbclass
+++ b/classes/shared-source-use.bbclass
@@ -1,4 +1,3 @@
 S = "${TMPDIR}/work-shared/${SOURCE_NAME}-${PV}-${PR}"
 
-do_configure[depends] += "${SOURCE_NAME}-source-${PV}:do_patch"
-do_populate_lic[depends] += "${SOURCE_NAME}-source-${PV}:do_unpack"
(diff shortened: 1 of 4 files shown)


jmesmon:
Why not simply patch all of the source via the shared source mechanism? (instead of patching separately for llvm).

derekstraka:
I agree with using the shared source.  I made that change.

In the environment I was using to test, all the build deps for rust were already in the SSTATE, so nothing else had to be built, and I ran into an error

The first warning from do_unpack complaining about the S dir not existing and then the function
do_rust_arch_fixup can failed since it was in the middle of a fetch when it executed.

jmesmon:
Thanks, merged.

PR #42, opened 2016-02-19
Remove the '--disable-static' in the EXTRA_OECONF for all rust-like recipes

### LINKED ISSUE #40: rust-cross-${ARCH} fails to build with latest poky
With poky after [438d6d6e7d8ca1a9d566bbea5b917a2b00a31f4e](https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=438d6d6e7d8ca1a9d566bbea5b917a2b00a31f4e) poky defines an EXTRA_OECONF_append to '--disable-static' causing the following back trace in rust-cross
> 
| configure: calculating relative paths to prefix = /var/build/build/tmp/sysroots/x86_64-linux/usr
| configure:
| configure:
| configure: validating /var/build/build/tmp/work-shared/rust-1.5.0-r0/rustc-1.5.0/configure args
| configure:
| configure: error: Option '--disable-static' is not recognized
| WARNING: exit code 1 from a shell command.
| ERROR: Function failed: do_configure (log file is located at /var/build/build/tmp/work/x86_64-linux/rust-cross-x86_64/1.5.0-r0/temp/log.do_configure.11962)
ERROR: Task 134 (/var/build/poky/meta-rust/recipes/rust/rust-cross.bb, do_configure) failed with exit code '1'
> 

The OECONF needs to be filtered or a DISABLE_STATIC_pn-rust-cross-${ARCH} will need to be added.

As a work around the file poky/meta/conf/distro/include/no-static-libs.inc can be modified to remove the extra append.

derekstraka:
Remove the newly introduced (in poky) --disable-static which is not recognized by the rust configure scripts.

Fixes #40

### DIFF
--- a/classes/rust.bbclass
+++ b/classes/rust.bbclass
@@ -82,3 +82,5 @@ HOST_LDFLAGS  ?= "${LDFLAGS}"
 HOST_CFLAGS   ?= "${CFLAGS}"
 HOST_CXXFLAGS ?= "${CXXFLAGS}"
 HOST_CPPFLAGS ?= "${CPPFLAGS}"
+
+EXTRA_OECONF_remove = "--disable-static"


jmesmon:
Thanks.

PR #44, opened 2016-03-18
Update rust to version 1.7.0

### LINKED ISSUE #27: update to rust-1.7.0

derekstraka:
resolves #27


jmesmon:
I'm in the process of testing this out.

1 question: why is the addition of libssp to the post link flags necessary with this update to rust? What happens when it is omitted? Is it something we'll always need, or is it potentially distro dependent?

derekstraka:
Without the libssp, the linking fails with undefined references to stack_chk_fail

derekstraka:
I can add SECURITY_LDFLAGS_append locally for ARM if people aren't using the stack protection generally and just pop that commit from the PR

jmesmon:
This appears to work OK for me with or without the `-lssp` addition.

I'd prefer both not adding it all the time if not everyone requires it, and if it's something that people are going to commonly run into, I'd prefer we support it somehow within meta-rust itself (so folks don't need to go hunting around too much for a solution.

Perhaps it could be added to an include like (or even in the same file as) `conf/distro/include/rust_security_flags.inc`? If not, I suppose a `PACKAGECONFIG` value would also work.

derekstraka:
Added the LD options to `conf/distro/include/rust_security_flags.inc`

jmesmon:
Works for me, thanks again for putting this together.

PR #45, opened 2016-04-06
Add libssh2 recipe from openembedded

cajun-rat:
The libssh2 recipe isn't in Poky/Jethro. Import it into meta-rust so meta-rust
only depends on Poky.

This code was taken from the jethro branch of poky, commit ID
c305ac5d2f5285d5eec8952a4ca7f3b4f89aed96

### DIFF
--- a/recipes-deps/libssh2/libssh2_1.6.0.bb
+++ b/recipes-deps/libssh2/libssh2_1.6.0.bb
@@ -1,0 +1,24 @@
+SUMMARY = "A client-side C library implementing the SSH2 protocol"
+HOMEPAGE = "http://www.libssh2.org/"
+SECTION = "libs"
+
+DEPENDS = "zlib"
+
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://COPYING;md5=c5cf34fc0acb44b082ef50ef5e4354ca"
+
+SRC_URI = "http://www.libssh2.org/download/${BP}.tar.gz"
+SRC_URI[md5sum] = "00aabd6e714a5f42a4fb82ace20db1dd"
+SRC_URI[sha256sum] = "5a202943a34a1d82a1c31f74094f2453c207bf9936093867f41414968c8e8215"
+
+inherit autotools pkgconfig
+
+EXTRA_OECONF += "\
+                 --with-libz \
+                 --with-libz-prefix=${STAGING_LIBDIR} \
+                "
+
+# only one of openssl and gcrypt could be set
+PACKAGECONFIG ??= "openssl"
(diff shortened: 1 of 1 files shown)


cajun-rat:
I've kept the bbappend and the original recipe separate, so that the BBCLASSEXTEND = "native" will still be applied if a user pulls in a layer that has a newer version of libssh2.

jmesmon:
Well, I'm a bit iffy on including this, but if it is the only thing from meta-openembedded we're using, it might make sense to do.

If we're going to include it ourselves, it probably would make sense to use `libssh-1.7.0`, which has a fix for (CVE-2016-0787)[https://www.libssh2.org/adv_20160223.html] (the bug causes the security from DH to be reduced to almost nothing).

posborne:
I'm not in favor of including recipes from other layers except at the top-level (the layer for your actual target).  That isn't the case here, so collapsing in recipes from meta-openembedded/* seems like a Bad Idea.  There is a very good chance that my project will need other things from meta-openembedded/* and then the user has a conflict.

@cajun-rat If you haven't already done so, I would encourage you to create a new layer for your project and add the recipe there if you don't want to include meta-openembedded for some reason (adding a layer itself does not typically cause significant problems).

derekstraka:
We also pull in functionality from meta-openembedded for our projects.  I'd prefer leaving the recipe where it is to avoid duplicate recipe issues.

cajun-rat:
I'm not familiar enough with Yocto to know to to work around duplicate recipe conflicts: is there a suffix like Debian's ~prerelease that causes a recipe to be considered just a little be earlier than the 'normal' recipe? That would avoid recipe conflicts for people who already depend on openembedded, and also work out of the box for people who just have a bare set of Poky layers. 

There's no problem from my side moving the recipe to our own layer,  but it feels a little bit nicer for new users if they don't need to sort out the dependency themselves.

cajun-rat:
And thank you for the heads up on the security bug.

posborne:
@cajun-rat Check out the following.

* [PREFERRED_PROVIDER](http://www.yoctoproject.org/docs/2.0.1/mega-manual/mega-manual.html#var-PREFERRED_PROVIDER)
* [PREFERRED_VERSION](http://www.yoctoproject.org/docs/2.0.1/mega-manual/mega-manual.html#var-PREFERRED_VERSION)

Conflicts in recipes will be resolved by the layer priority.  Typically, your most specific layer (for your project/board) will have the highest priority.  See http://www.yoctoproject.org/docs/1.8/dev-manual/dev-manual.html#prioritizing-your-layer

PR #46, opened 2016-04-11
Fix rust build

cardoe:
This work brings our x86_64 build inline with what official Rust releases produce as far as code generation goes.


jmesmon:
Thanks. And I think we want all of this (so I've merged it).

A question for posterity:

 - what effect will removing `--enable-debuginfo` have on the backtraces from `RUST_BACKTRACE=1` or the functionality of `rust-gdb` when used on a binary built with this `rustc` and `std`? Will we just not get info for the innards of `std` used in those programs?

On the rustc snapshot download issue:

I made the same mistake previously in cargo, and ended up adding a cargo patch to disable it's auto-download mechanism. In the long run it might make sense to provide some configurable option for cargo & rustc to prevent them from even trying to fetch.

cardoe:
So I dug in a bit and it looks like some conversation @derekstraka and I had offline about `--enable-debuginfo` was wrong and it appears to just build `rustc` with -g and has no effect on `std`. I'll look to see if the official builds are done that way because RUST_BACKTRACE=1 works fine with official builds so if they don't need it then we shouldn't either.

As far as the snapshot bits, I agree with you some switch would be best. I've got the same issue in Gentoo... https://github.com/gentoo/gentoo/commit/f3c24321c819a2bccd4e9f944260bf77b3ad5140 Probably should file a bug upstream.

PR #50, opened 2016-05-06
Several fixes for rust 1.7

srwalter:
Most importantly, update the stage0 snapshot so that it actually accomplishes something.  Additionally, improve support for running on tiny Linux systems with support for musl libc and static linking.


jmesmon:
looks good to me, thanks!

PR #51, opened 2016-05-11
Replace rust-cross with a cross-compiled libstd

srwalter:
Posting this more for discussion than immediately merge.

Properly cross-compiling libstd is much faster than building the entire toolchain again, plus it should behave more nicely re: sstate hashes.  This way you only have to build the rust compiler once for a given build OS, and can quickly cross-compile for arbitrary target OSes.  There was already a comment to the effect that we should directly cross-compile libstd and friends, and now we do.

The reason I'm not proposing to immediately merge this is that I suspect the libstd-rs library is fragile and will break on future rust upgrades.  Rust 1.8.0 already looks to use cargo to build libstd, so that should be a much more sustainable approach.  I'll rebase these patches once meta-rust supports 1.8.0


jmesmon:
Nice!

Looks like this doesn't let us parallelize building the crates, but I think it's still a step in the right direction.

Note that we probably need to adjust `rust.bbclass`'s deps.

aside: rust.bbclass was originally indented as the class for packages that were written in rust to inherit, so it might make sense to merge rust-bin.bbclass and rust.bbclass. `rust.bbclass` does have a bunch of cruft in it from back when I was still trying to decide whether to use only pre-supplied triples or generate our own, some of it can probably be moved into compiler-only include or bb files.

srwalter:
Yeah I think holding off until 1.8.0 is a good idea.  Might also make sense to get @tylerwhall's cargo support in first, as well.

srwalter:
I've updated this now that we're on rust 1.10 and can use cargo for building libstd.  Please look this over again; if there's no objection I feel confident in merging this version.

jmesmon:
I haven't had a chance to test, but it looks good.

A few questions:

 - is there any chance that libbacktrace.a or libcompiler-rt.a could conflict with something else installed in `${libdir}`? (not sure what the status is of libs with those names outside of rustc).

 - The usage of `${prefix}/lib` for target spec generation in some of the recipes strikes me as probably not quite right. We might want to use `${prefix}/${baselib}` instead, though I'm not sure what the exact rules are for this in bitbake.

 - For 0012-libstd-don-t-build-libbacktrace.patch I believe we'll want to upstream a way to disable building (probably an env var checked in it's build script).

srwalter:
> is there any chance that libbacktrace.a or libcompiler-rt.a could conflict with something else installed in ${libdir}? (not sure what the status is of libs with those names outside of rustc).

libbacktrace and compiler-rt are both open-source projects independent of rust, so I would not expect namespace collision to be a problem.

> The usage of ${prefix}/lib for target spec generation in some of the recipes strikes me as probably not quite right. We might want to use ${prefix}/${baselib} instead, though I'm not sure what the exact rules are for this in bitbake.

Good call, I've made that change.

> For 0012-libstd-don-t-build-libbacktrace.patch I believe we'll want to upstream a way to disable building (probably an env var checked in it's build script).

Fortunately, I was able to drop this patch.  The reason libbacktrace was giving me trouble was because oe_cargo_fix_env was missing from the recipe for libstd-rs.  I added it and now libstd-rs can build libbacktrace itself.

Assuming there's no further concerns, I plan to merge this latest version sometime this weekend

Cheers!

PR #52, opened 2016-05-28
Integrate Cargo into the Bitbake package system

tylerwhall:
This lays the foundation to move away from raw rustc in the rust-bin bbclass to being able to build with cargo.

- Cargo bbclass cleanups and sharing features with rust-bin
- Cargo builds without Internet access
- Cargo-based recipes publish their crate to a local sysroot registry
- Cargo-based recipes reference a local registry, downloading crates from the sysroot
- Cargo can be configured to only build the root crate and no dependencies, looking for the dependencies in the sysroot (the default for cargo bbclass)

I'm not converting any recipes to cargo in this set. I have tested converting libc and a couple test crates, but there is more work to do to resolve undeclared dependencies and compatible versions in the set of libraries already in this layer.


jmesmon:
This looks like a good direction to proceed in. Getting no network builds working with cargo would be really nice.

I'm a bit leary about us adding the bunch of generated json files (cargo_index), but it seems like it should be possible to move away from that in the future if someone is interested in doing so.

@srwalter @cardoe @derekstraka  any issues with these changes?

srwalter:
Looks good to me.  I much prefer being able to use the "native" build system for recipes instead of having to manually invoke rustc

tylerwhall:
Cargo wants to download a registry index even if all of its dependencies are satisfied by local paths in .cargo/config. It would be possible to extract the crates and generate the json registry files during the build, but there's not a direct cargo command to generate exactly the json required. However, the task in cargo.bbclass for publishing to the local sysroot registry contains python code to do it, so that could be re-used.

classes/cargo.bbclass
```diff
@@ -84,11 +94,87 @@ cargo_do_compile () {
...
+
+    missing = set(wanted_keys) - set(pkg)
+    if missing:
+        raise bb.build.FuncFailed("Missing keys %s in pkg manifest" % missing)
+
+    name = pkg["name"]
+    if len(name) == 1:
+        path = os.path.join("1")
+    elif len(name) == 2:
+        path = os.path.join("2")
```
tylerwhall: Unnecessary os.path.join.

posborne:
I'm testing right now.  Off the bat, ran into this.  I've seen this before and the solution is to not use pushd (unfortunately):

```
| /opt/builds/paul-hubv2/build/tmp/work/x86_64-linux/cargo-native/0.10.0-r0/temp/run.do_cargo_local_registry.13045: 113: /opt/builds/paul-hubv2/build/tmp/work/x86_64-linux/cargo-native/0.10.0-r0/temp/run.do_cargo_local_registry.13045: pushd: not found
```

I'll fix locally and keep trucking and let you know how things go.

cardoe:
A lot to review. I'll try to get started on it over the next few weeks. I'm in the process of getting cargo packaged up nicely under Gentoo. I'm fetching down the crates and the index manually and reassembling what cargo expects it would need since Gentoo unfortunately doesn't have a fetch phase where I can just call cargo. I figured it would be easier to land this working in Gentoo properly than under BitBake. Once I finish the Gentoo bits I'll turn my attention back here. For reference: https://github.com/cardoe/gentoo/tree/rust-cargo

cardoe:
So I'll say I'm not a fan of breaking each Rust crate as its own BitBake recipe. Its going to make maintenance harder since there are a number of recipes that will have to be updated in each go and we'll have to have virtually every version of crates packaged up as well.

Additionally another issue I see is fetching the crates from their upstream repo. The tarballs that you're getting based on GitHub's API aren't necessarily the same tarball that's uploaded to crates.io. A number of upstreams also don't publish tags for their crates.

tylerwhall:
The scalability of having a recipe for each crate is a concern. I'm hopeful that we can find a way to automate the maintenance of those, or even auto-generate the recipe data at build time since cargo is so standardized. But without exposing each crate individually to bitbake, I don't know of a way to benefit from shared state and, more importantly for embedded, shared libraries.

Internally cargo is flexible in finding a common compatible version of shared dependencies within a project, even if different semvers are specified explicitly in multiple Cargo.tomls. I'm hoping this will afford us some flexibility in not having to have every possible library version in the build. There are some early parsing stages in cargo where it is not flexible which will have to be addressed, but for now the goal of this patch set it to make it at all possible to share libraries between crates.

cardoe:
Well these aren't shared libraries. They're effectively static libraries that are linked into the final binary. You're saving yourself a little bit of time on the compilation at the expense of a lot of maintenance.

Another potential issue I see is that crates have features and different crates depend on other crates being built with a feature on or off.

cardoe:
I've implemented a crate fetcher (crate://NAME/VERSION) similar to how I did the Gentoo bits. You can see an example of a package here: https://github.com/starlab-io/meta-rust/blob/crate-fetch/recipes-example/rustfmt/rustfmt_0.4.0.bb The branch is not complete yet but it does not reach out to the network and is much simpler than the current cargo support in the repo. I'm also working on a cargo extension which can generate these bitbake recipes for packages.

jmesmon:
@cardoe that looks really nice, and being able to automate generating it from cargo metadata seems like it would make the perfect solution (well, baring having bitbake somehow directly understand Cargo.toml files & crates.io indexes, and I'm pretty sure that'd be quite a bit of work and might not even be possible within the model of bitbake).

tylerwhall:
@cardoe the fetcher class seems nice for streamlining self-contained static builds. That's a use-case I think we should support by default since it's the expected cargo developer workflow and should be a quick path to shippable cargo no-network builds. I'm a little wary of relying on the implementation details of cargo's on-disk cache, but it is more efficient than implementing the public crates.io API so I'm not partial to either approach. Hopefully cargo will eventually get first-class support for a local registry. https://github.com/rust-lang/cargo/issues/2212

I like this better than what I did to get cargo-native bootstrapped, but that wasn't the focus of this patch set. I'm still working on the orthogonal problem of making cargo work cooperatively with the distro's package manager. rustc is perfectly capable of generating incrementally built shared objects which are common between multiple apps. This is how the current set of raw rustc recipes in this layer are structured. The goal is not primarily build-time efficiency but disk space and RAM efficiency on the target when a large number of rust apps are installed. I'm thinking of an end-user which may have a number of their own apps that they control, so maintaining compatible dependency versions between them is not so much of an issue.

This patch set can be rebased to take advantage of the crate fetcher if that's merged, but one does not preclude the other since they're solving different problems for different end use-cases.

tylerwhall:
I'll also caution that the use of a full registry git snapshot gives cargo the possibility to fail by selecting a different version at build time than when the recipe was generated. This shouldn't happen if all versions including cargo-native are exactly the same as when the recipe was generated, but that means that any tweak to the recipe like overriding dependencies (which is done in cargo-native) or a change to the version of cargo-native should require the recipe to be regenerated. The registry metadata could be generated from the contents of the crates to make it resilient to minor changes.

cardoe:
> I'll also caution that the use of a full registry git snapshot gives cargo the possibility to fail by selecting a different version at build time than when the recipe was generated. This shouldn't happen if all versions including cargo-native are exactly the same as when the recipe was generated, but that means that any tweak to the recipe like overriding dependencies (which is done in cargo-native) or a change to the version of cargo-native should require the recipe to be regenerated. The registry metadata could be generated from the contents of the crates to make it resilient to minor changes.

The full git snapshot is actually purposeful. I've written a command called `cargo-bitbake` that you install like `rustfmt` (which gives `cargo fmt`) and yields a `cargo bitbake` command. It takes a project and generates a `Cargo.lock` and using the revision of the registry that the `Cargo.lock` references it lists out all the dependencies so nothing should change. I'm working on adding different arches in but there should never be a reason to doctor up the dependencies like cargo-native. The main difference is that BitBake files won't really be user editable to change versions but instead will be generated each time. We could certainly generate a custom registry index for each package with `cargo-vendor` but that would be way too much work than its worth.

tylerwhall:
I'm still ironing out a few minor issues with Cargo's behavior in trying to build a large variety (>100) of crates using this technique, so it's become clear that this isn't ready for real-world use in its current state. I'll see if I can split up the bbclass cleanups into a separate branch for merging soon, and then I'd like to rebase this feature onto the crate fetcher when that's available.

Issue #53, opened 2016-06-21
failure in docker


kteza1:
Hi. I'm trying to use meta-rust in a docker container. `bitbake rust-hello-world` is throwing this error

```
x86_64-linux/stage1/bin/arm-poky-linux-gnueabi/rustc: symbol lookup error: /home/build/buildbot/slave-jethro/jeth/main/build/tmp/work/x86_64-linux/rust-cross-a
rm/1.7.0-r0/build/x86_64-linux/stage1/lib/arm-poky-linux-gnueabi/libterm-6a154fe0.so: undefined symbol: _ZN2io5error18Error.error..Error11description20h97638b304e7c4d084tgE

| x86_64-linux/stage1/bin/arm-poky-linux-gnueabi/rustc: symbol lookup error: /home/build/buildbot/slave-jethro/jeth/main/build/tmp/work/x86_64-linux/rust-cross-arm/1.7.0-r0/build/x86_64-linux/stage1/lib/arm-poky-linux-gnueabi/libterm-6a154fe0.so: undefined symbol: _ZN2io5error18Error.error..Error11description20h97638b304e7c4d084tgE
```
Any insights?

cardoe:
We use this in a docker container always. Docker isn't the issue here. What's the distro that your container is based on? @derekstraka have we seen something like this? I vaguely remember something like that.

kteza1:
@cardoe I tried both ubuntu 14.04 & 15.10 containers. (but it works properly outside container on my ubuntu 15.10 machine)

derekstraka:
@cardoe - we did see an error similar to this.  I don't recall if we just removed our SSTATE_DIR and it built or if I added a fix in our fork.  

@RT - Any chance you could try https://github.com/starlab-io/meta-rust/ to see if it builds?  If it does, that will give us an idea of what may be the underlying issue.

kteza1:
@derekstraka @cardoe Oh yeah. `libterm` undefined symbol error is gone with `starlab-io/meta-rust`. But unfortunately, I'm facing the below error [which I previously reported](https://github.com/meta-rust/meta-rust/issues/49) (this was a problem in fido but not jethro)

```
cargo: /home/build/buildbot/slave-jeth/jeth/main/build/tmp/sysroots/x86_64-linux/usr/bin/cargo
[...]

PR #56, opened 2016-06-30
libgit2: add a dependency on openssl

cardoe:
If this dependency is missing its possible that openssl won't be built
before libgit2 is built and then we won't have git support over SSL
which leaves out a lot of git repos. Should fix the issue mentioned in #53.


cardoe:
If this is right I want to backport this to the jethro branch.

derekstraka:
This looks good to me.  For some reason I thought libgit2 was in meta-oe and not just in meta-rust

derekstraka:
@srwalter @jmesmon Any concerns here before I merge

jmesmon:
nope, looks good. 👍

Issue #48, opened 2016-04-19
Upgrade to 1.10 stable


realglotzi:
There is a new stable version 1.8 on the road and it would be great to have an update of meta-rust.

alexhumphreys:
Any news on this issue?

kteza1:
Any plans for 1.10 soon?

hmvp:
I tried upgrading but between my limited knowledge of the rust build system and issue #47 I could not make it work yet
Here is what I tried: https://github.com/hmvp/meta-rust/tree/try-upgrade-rust

posborne:
I am also very interested in moving to a newer version.  @hmvp I tried your branch a few days ago but ran into some other issues when building the rustlib recipe and haven't been able to fully debug as of yet.

cardoe:
Honestly your issue is likely related to the change of the target triples. We need to move to the standard ones and away from using the Yocto ones. I made the change in [Gentoo Linux](https://github.com/gentoo/gentoo-rust/blob/master/dev-lang/rust/) with less code and less hassle between upgrades but I haven't been successful with Yocto.

cardoe:
I should also mention that I've been working on trying to push changes we have in this layer upstream which means the changes are currently landing for 1.12 so I think we end up targeting 1.10 for a krogoth branch and potentially 1.12 for master. 

/cc @derekstraka

PR #57, opened 2016-07-11
Update Cargo, bbclass cleanup

tylerwhall:
Move some variables from `rust-bin.bbclass` to `rust.bbclass`, working toward commonality between rust-bin and cargo.

The new version of cargo allows a couple `cargo.bbclass` hacks to be removed.

I tested that dynamically-linked rust-bin apps execute on the generated target image and rust-hello-world (cargo) builds.

### DIFF
--- a/classes/rust.bbclass
+++ b/classes/rust.bbclass
@@ -15,7 +15,7 @@ def rust_base_dep(d):
             deps += " rust-native"
     return deps
 
-DEPENDS_append = " ${@rust_base_dep(d)}"
+DEPENDS_append = " ${@rust_base_dep(d)} patchelf-native"
 
 def rust_base_triple(d, thing):
     '''
@@ -85,6 +85,21 @@ HOST_CXXFLAGS ?= "${CXXFLAGS}"
 HOST_CPPFLAGS ?= "${CPPFLAGS}"
 
 EXTRA_OECONF_remove = "--disable-static"
+
+do_rust_bin_fixups() {
+    for f in `find ${PKGD} -name '*.so*'`; do
+        echo "Strip rust note: $f"
+        ${OBJCOPY} -R .note.rustc $f $f
+    done
+
+    for f in `find ${PKGD}`; do
+        file "$f" | grep -q ELF || continue
+        readelf -d "$f" | grep RUNPATH | grep -q rustlib || continue
+        echo "Set rpath:" "$f"
+        patchelf --set-rpath '$ORIGIN:'${rustlibdir}:${rustlib} "$f"
+    done
+}
+PACKAGE_PREPROCESS_FUNCS += "do_rust_bin_fixups"
 
 rustlib_suffix="${TUNE_ARCH}${TARGET_VENDOR}-${TARGET_OS}/rustlib/${HOST_SYS}/lib"
 # Native sysroot standard library path
--- a/classes/rust.bbclass
+++ b/classes/rust.bbclass
@@ -3,13 +3,14 @@ RUSTC = "rustc"
 # FIXME: --sysroot might be needed
 RUSTC_ARCHFLAGS += "--target=${TARGET_SYS} -C rpath -C crate_hash=${BB_TASKHASH}"
 
+RUSTLIB_DEP ?= "rustlib"
 def rust_base_dep(d):
     # Taken from meta/classes/base.bbclass `base_dep_prepend` and modified to
     # use rust instead of gcc
     deps = ""
     if not d.getVar('INHIBIT_DEFAULT_RUST_DEPS', True):
         if (d.getVar('HOST_SYS', True) != d.getVar('BUILD_SYS', True)):
-            deps += " virtual/${TARGET_PREFIX}rust"
+            deps += " virtual/${TARGET_PREFIX}rust ${RUSTLIB_DEP}"
         else:
             deps += " rust-native"
     return deps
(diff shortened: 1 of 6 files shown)


jmesmon:
looks good to me

recipes-devtools/cargo/files/0001-disable-cargo-snapshot-fetch.patch
```diff
@@ -8,20 +8,20 @@ Subject: [PATCH] disable cargo snapshot fetch
...
-@@ -92,10 +92,6 @@ test-unit-$(1): $$(CARGO)
+@@ -103,10 +103,6 @@ test-unit-$(1): $$(CARGO)
  endef
  $(foreach target,$(CFG_TARGET),$(eval $(call CARGO_TARGET,$(target))))
  
--$(TARGET_ROOT)/snapshot/bin/cargo$(X): src/snapshots.txt
--	$(CFG_PYTHON) src/etc/dl-snapshot.py $(CFG_BUILD)
+-$(TARGET_ROOT)/snapshot/bin/cargo$(X): $(S)src/snapshots.txt
+-	$(CFG_PYTHON) $(S)src/etc/dl-snapshot.py $(CFG_BUILD)
 -	touch $@
```
cardoe: This snippet landed in cargo 0.10.0 so it should be unnecessary. I just checked the 0.10.0 tarball and see the following:

```
$(TARGET_ROOT)/snapshot/bin/cargo$(X): $(S)src/snapshots.txt
    $(CFG_PYTHON) $(S)src/etc/dl-snapshot.py $(CFG_BUILD)
    touch $@
```

classes/rust.bbclass
```diff
@@ -86,6 +86,21 @@ HOST_CPPFLAGS ?= "${CPPFLAGS}"
 
 EXTRA_OECONF_remove = "--disable-static"
 
+do_rust_bin_fixups() {
```
cardoe: So the goal here is to set the rpath to the proper rustlib path? I thought some handling landed in rust to automagically set the proper rpaths during build time and installation time. @jmesmon is my understanding of this wrong? This comment shouldn't block merging this series in but more of a clarification.

cardoe:
Two things of note:

1. This series looks like its done from the point of "bump cargo... now what breaks?" So that will lead to a non-bisectable chunk of patches.
2. Rust and Cargo really go hand in hand together currently. While its not technically required it is recommended and this gets us the version of Cargo for Rust 1.9.0 but keeps the tree at Rust 1.7.0.

recipes-devtools/cargo/files/0001-disable-cargo-snapshot-fetch.patch
tylerwhall: This is rebasing the patch that removes those lines. I assume we still need to remove those since we're fetching the snapshot with bitbake, but admittedly I haven't looked into the details of how that's set up. Perhaps it could be removed as a separate change.

tylerwhall:
> that will lead to a non-bisectable chunk of patches.

This is bisectable. The ssh2-rs upgrade is independent of the cargo upgrade.

>Rust and Cargo really go hand in hand together currently

The last patch here relies on RUSTFLAGS which the later Cargo version provides. If someone is working on a compiler upgrade that will land soon (#48), we could wait on this.

classes/rust.bbclass
tylerwhall: The observed behavior is that the compiler inserts a runpath when appropriate but it's a relative path from the workdir to the sysroot. I haven't personally looked at the underlying logic that chooses the path.

It's not pretty, but for now we just override it for any ELFs for which the compiler set a runpath.

Before:

`Library runpath: [$ORIGIN/../../../../../sysroots/qemux86/usr/lib/rust:$ORIGIN/../../../../../sysroots/qemux86/usr/lib/i586-poky-linux/rustlib/i586-poky-linux/lib:/<ABSOLUTE PATH TO TOPDIR>/tmp/sysroots/x86_64-linux/usr/lib/rustlib/i586-poky-linux/lib]`

After patchelf:

`Library runpath: [$ORIGIN:/usr/lib/rust:/usr/lib/i586-poky-linux/rustlib/i586-poky-linux/lib]`

And stripping the .note.rustc is done because the section is huge and not needed on the target, but doesn't get pulled out by strip by default.

recipes-devtools/cargo/files/0001-disable-cargo-snapshot-fetch.patch
jmesmon: The removal isn't strictly required, but it does make sure that we actually grab the snapshot ourselves instead of having cargo's build download it behind our backs (which has happened a few times in the past when our updates weren't quite right)

cardoe: Ah you're right. I didn't follow the double minuses and double pluses properly. However as long as we `touch` the cargo snapshot binary this rule shouldn't get called. I'd personally rip out the patch.

srwalter:
Looks good to me, I'll merge later today if there's no objections.

cardoe:
Well I tried to build this branch and I get a number of failures. But they're the same failures I get with master currently.

I however cherry picked these patches on top of the jethro branch and see the following:
```
ERROR: QA Issue: cargo: Files/directories were installed but not shipped in any package:
  /usr/lib
  /usr/etc
  /usr/lib/rustlib
  /usr/lib/rustlib/manifest-cargo
  /usr/share/zsh
  /usr/share/zsh/site-functions
  /usr/share/zsh/site-functions/_cargo
  /usr/etc/bash_completion.d
  /usr/etc/bash_completion.d/cargo
Please set FILES such that these items are packaged. Alternatively if they are unneeded, avoid installing them or delete them within do_install.
cargo: 9 installed and not shipped files. [installed-vs-shipped]
```

I don't see any handling in master for these files.

cardoe:
Let's merge and improve bits going forward.

cardoe:
This actually forgot to change the Cargo snapshot that is used to build Cargo. PR coming to fix that.

PR #58, opened 2016-07-12
add a custom fetcher for Cargo crates

### LINKED ISSUE #12: Consider having the entire crates.io index/db automatically imported as bb packages
This would also include adding a crates.io fetcher to get the source code for a crate.

### LINKED ISSUE #24: add fetcher support for crates.io
Should be similar to the npm fetcher implimented in bitbake-npm.

Would allow us to avoid using git sources when patching up cargo, and could make it much more convenient to build some packages.

cardoe:
This adds a custom fetcher for Cargo crates and hopefully will simplify adding Cargo based projects to meta-rust. The URL syntax is:

`crate://HOST/NAME/VERSION` with a standard example being `crate://crates.io/memchr/0.1.11`

Additionally this supports fetching down a snapshot of the registry index for reproducible builds with the syntax:

`crate-index://HOST/REVISION` with an example being `crate-index://crates.io/3b3994e099281c394a6a66604d1af6c0920e4c31`

`crates.io` as a HOST magically fixes up the paths to hit the official registry. This should be able to support custom registries should people want to use them as well.

Fixes #24. Might close #12.

### DIFF
--- a/classes/cargo_util.bbclass
+++ b/classes/cargo_util.bbclass
@@ -1,0 +1,87 @@
+# add crate fetch support
+inherit crate-fetch
+
+# the binary we will use
+CARGO = "cargo"
+
+# Where we download our registry and dependencies to
+export CARGO_HOME = "${WORKDIR}/cargo_home"
+
+# We need cargo to compile for the target
+BASEDEPENDS_append = " cargo-native"
+
+# Ensure we get the right rust variant
+DEPENDS_append_class-target = "virtual/${TARGET_PREFIX}rust"
+
+# Cargo only supports in-tree builds at the moment
+B = "${S}"
+
+# In case something fails in the build process, give a bit more feedback on
+# where the issue occured
+export RUST_BACKTRACE = "1"
+
+# The pkg-config-rs library used by cargo build scripts disables itself when
+# cross compiling unless this is defined. We set up pkg-config appropriately
+# for cross compilation, so tell it we know better than it.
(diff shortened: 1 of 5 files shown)


derekstraka:
I built rustfmt against aarch64 and it was successful

cardoe:
@jmesmon @tylerwhall @srwalter Would you guys like to review this?

jmesmon:
It looks fine to me.

I'd like some of the duplication between cargo_util.bbclass and cargo.bbclass to go away at some point, but I don't think that should necessarily be a blocker.

Might want to consider just having one bbclass for using cargo (could make it easier for folks), but I suppose that might be complicated by different people having different ways they want to use it.

cardoe:
That's my plan going forward @jmesmon. I'd like to drop cargo.bbclass and just have cargo_util.bbclass around. I just didn't want to break people immediately until I had https://github.com/cardoe/cargo-bitbake working fully.

srwalter:
:+1:

tylerwhall:
Looks good to me overall. I'll try to give it a test but don't wait for me to merge.

Seems like this would work well with a local mirror for the crates. I'm wondering if it would be better to fetch the index with git rather than http for the purpose of mirroring, though. I believe a set of recipes that were each generated against a different index rev would require a tarball of each of those snapshots to be individually mirrored. Using the git fetcher should mean you only need to keep one updated git tarball of the index.

We could save that for an enhancement if it's feasible, but it would slightly impact people who use local mirrors to change how it works.

PR #64, opened 2016-08-08
rustc: fix use of shared source for new poky/oe-core which wipes ${S}

jmesmon:
The oe-core/poky change 'base.bbclass wipe ${S} before unpacking
source' (poky: a56fb90dc3805494eeaf04c60538425e8d52efc5, oe-core:
eccae514b71394ffaed8fc45dea7942152a334a1) wipes ${S} in do_unpack prior
to unpacking. This breaks our shared source as we set ${S} to the shared
location, and don't actually unpack anything (the result is we try to
build rustc without any source code, which fails predictably)

Avoid this by copying a bit more of gcc-shared-source.inc to disable
do_unpack for shared source users.

This is a bit of a hack, and only barely works (due to our need for
rust-snapshot). In the future, it might make sense to have the
rust-snapshot be shared in some fashion as well (which would allow us to disable fetch and more closely mimic gcc-shared-source.inc).

### DIFF
--- a/classes/shared-source-use.bbclass
+++ b/classes/shared-source-use.bbclass
@@ -1,3 +1,23 @@
+# gcc's shared source code disables fetch (via the commented code below). We
+# can't do that because rust.bb currently fetches a rustc-snapshot archive itself.
+#do_fetch() {
+#	:
+#}
+#do_fetch[noexec] = "1"
+
+# gcc does `deltask` do_unpack. We avoid this so that the depends work sanely
+# (things that need source code can still be ordered after do_unpack).
+# As a side effect, we can also unpack things that aren't shared.
+# Note: just setting this normally doesn't work. Use of python() is required.
+python () {
+	d.setVarFlag('do_unpack', 'cleandirs', '')
+}
+# Avoid disabling do_patch for the same reason.
+#deltask do_patch
+
+SRC_URI = ""
+
 S = "${TMPDIR}/work-shared/${SOURCE_NAME}-${PV}-${PR}"
 
 do_unpack[depends] += "${SOURCE_NAME}-source-${PV}:do_patch"
+do_populate_lic[depends] += "${SOURCE_NAME}-source-${PV}:do_unpack"


jmesmon:
@srwalter @cardoe @derekstraka ^

[I'd do @meta-rust , but I don't think that works. Might put everyone in a team so group notifications are possible]

srwalter:
I agree that this is a bit hacky, but it looks like it should work.  I also agree that moving all the fetching into the shared-source recipe would probably help clean things up

jmesmon:
imho, even the gcc-shared-source bits are fairly hacky. But they're less likely to be broken due to more things depending on gcc :) 

AFAIK, gcc is the only package with a similar shared source setup.

derekstraka:
@jmesmon - Were you able to get this to build using meta-oe master?  I'm hitting issues in rust_gen_mk_cfg

derekstraka:
IIRC the kernel also puts the source in work-shared.

jmesmon:
wrt kernel using work-shared: yep, they do. It looks like their recipies fully expect do_unpack() to clear `${S}` and `${STAGING_KERNEL_DIR}`.

It isn't really clear how they're handling actually having `${S}` be populated.

jmesmon:
@derekstraka my test here was with the versions I spec'd in #65 .

What error are you seeing?

derekstraka:
ERROR: rust-cross-x86_64-1.7.0-r0 do_rust_arch_fixup: Error executing a python function in exec_python_func() autogenerated:

The stack trace of python calls that resulted in this exception/failure was:
'exec_python_func() autogenerated', lineno: 2, function: <module>
     0001:
 *** 0002:do_rust_arch_fixup(d)
     0003:
'/var/build/poky/meta-rust/recipes-devtools/rust/rust.inc', lineno: 359, function: do_rust_arch_fixup
     0355:
     0356:python do_rust_arch_fixup () {
     0357:    for thing in ['BUILD', 'HOST', 'TARGET']:
     0358:        bb.debug(1, "rust_gen_mk_cfg for " + thing)
 *** 0359:        rust_gen_mk_cfg(d, thing)
     0360:}
     0361:addtask rust_arch_fixup before do_configure after do_patch
     0362:do_rust_arch_fixup[dirs] += "${WORKDIR}/mk-cfg"
     0363:
'/var/build/poky/meta-rust/recipes-devtools/rust/rust.inc', lineno: 320, function: rust_gen_mk_cfg
     0316:    ldflags = ' '.join(ldflags_for(d, thing, arch))
     0317:
     0318:    b = d.getVar('WORKDIR', True) + '/mk-cfg/'
     0319:    o = open(b + sys_for(d, thing) + '.mk', 'w')
 *** 0320:    i = open(d.getVar('S', True) + '/mk/cfg/' + rust_base_sys + '.mk', 'r')
     0321:
     0322:    r = subprocess.call(['sed',
     0323:        # update all triplets to the new one
     0324:        '-e', 's/{}/{}/g'.format(rust_base_sys, sys),
Exception: FileNotFoundError: [Errno 2] No such file or directory: '/var/build/build/tmp/work-shared/rust-1.7.0-r0/rustc-1.7.0/mk/cfg/x86_64-unknown-linux-gnu.mk'

jmesmon:
Hm... that does look like it's still removing '${S}'. What layers are in that setup? I'll try to reproduce it when I get a chance.

derekstraka:
I'm wondering it is due to the task order not being explicitly defined.  I'm using BB_NUMBER_THREADS=64 and running `bitbake rust-hello-world`

Just the typical suspects meta-oe, meta-rust, and meta-openembedded

POKY_BBLAYERS_CONF_VERSION = "2"

BBPATH = "${TOPDIR}"
BBFILES ?= ""

BBLAYERS ?= " \
  /var/build/poky/meta \
  /var/build/poky/meta-poky \
  /var/build/poky/meta-yocto-bsp \
  /var/build/poky/meta-openembedded/meta-oe \
  /var/build/poky/meta-openembedded/meta-networking \
  /var/build/poky/meta-openembedded/meta-python \
  /var/build/poky/meta-openembedded/meta-ruby \
  /var/build/poky/meta-rust \
  "
BBLAYERS_NON_REMOVABLE ?= " \
  /var/build/poky/meta \
  /var/build/poky/meta-poky \
  "

jmesmon:
> I'm wondering it is due to the task order not being explicitly defined

Ah yes, that sounds about right. Nothing causes `do_rust_arch_fixup` to run prior to source unpack. Must have just happened to work for me.

jmesmon:
Ok, I've tweaked this to instead just clear `do_unpack[cleandirs]` (and add some unimportant bits from `gcc-shared-source.inc`. Should resolve the ordering issue.

derekstraka:
That fixes the issue for me.  Thanks!

Issue #47, opened 2016-04-13
rust-cross build failure due to missing llvm symbols using `cxx11` references


jmesmon:
Doesn't happen on all systems. I've observed this on arch linux, but not on gentoo. Might depend on whether the system gcc has the new cxx11 abi enabled.

Probably due to `uninative.bbclass` from selectively disabling gcc's new cxx11 abi & `cross.bbclass` having a different `*_CXXFLAGS` compared to `native.bbclass`
```
# https://wiki.debian.org/GCC5
# We may see binaries built with gcc5 run or linked into gcc4 environment
# so use the older libstdc++ standard for now until we don't support gcc4
# on the host system.
BUILD_CXXFLAGS_append = " -D_GLIBCXX_USE_CXX11_ABI=0"
```

Full error output
```
| error: linking with `gcc` failed: exit code: 1
[...]

PR #67, opened 2016-08-28
Upgrade rust to 1.10.0

srwalter:
I had some trouble testing this against poky master due to uninative issues (I had the same problem without my changes).  However, I'm able to build rust-native, rust-cross-i586, and several rust packages using poky jethro.


jmesmon:
> I had some trouble testing this against poky master due to uninative issues

Are you referring to #47? If so, did you try the workaround I commented on here: https://github.com/meta-rust/meta-rust/issues/47#issuecomment-238139393 ?

jmesmon:
Patch looks good to me though.

tylerwhall:
rustc native/cross and cargo-native build for me on krogoth.

I should point out that Cargo 0.10 does not build with Rust 1.10, and the blessed corresponding Cargo, 0.11, does not build with Rust 1.7. So we have a bisectability issue with cargo-native, but it seems reasonable to me in this case to keep the commits logically separate.

Fixes #48

PR #71, opened 2016-09-03
Update cargo support so that there truly is no network access

srwalter:
We had done some preliminary work on this previously, but I doubt anyone had actually tested it in a firewalled environment before now.  With these changes we can build at least cargo-native and libstd-rs (which uses cargo built by cargo-native) in a fully firewalled build environment.  All fetching is done by bitbake during do_fetch in a mirror-friendly fashion.

Just to reiterate what I stated in a couple of the commit messages, there are still some ugly parts here due to the older version of cargo required by rust 1.10.  Once we get a new enough cargo, we can switch to using source-replacement entirely, which will allow a fair amount of clean-up.


jmesmon:
Nice.

So with these changes, cargo.bbclass (which presumes to fetch using cargo itself) is now fully broken, and folks are expected to use cargo_util.bbclass. Seems like it'd make sense to remove cargo.bbclass and potentially rename cargo_util.bbclass to it (or maybe not, while naming would improve, breaking compat would make a rename not so great).

Also, given that folks now have to figure out how to generate long lists of dependencies, it would be a good idea to update the README to point to @cardoe 's cargo-bitbake https://github.com/cardoe/cargo-bitbake so that it's not assumed to be a manual-only process.

On cargo + rustc versions: could you clarify what the restrictions are here that you're aware of?

srwalter:
> So with these changes, cargo.bbclass (which presumes to fetch using cargo itself) is now fully broken, and folks are expected to use cargo_util.bbclass. Seems like it'd make sense to remove cargo.bbclass and potentially rename cargo_util.bbclass to it (or maybe not, while naming would improve, breaking compat would make a rename not so great).

Good point.  I think on balance, it makes more sense to remove the old cargo.bbclass and replace it with cargo_util.bbclass.

> Also, given that folks now have to figure out how to generate long lists of dependencies, it would be a good idea to update the README to point to @cardoe 's cargo-bitbake https://github.com/cardoe/cargo-bitbake so that it's not assumed to be a manual-only process.

Agreed, I'll do that and update the request

> On cargo + rustc versions: could you clarify what the restrictions are here that you're aware of?

What I know for certain is that cargo 0.12 and newer cannot be built by rust 1.10.0.  There is a published toml file that lists the "sanctioned" version pairings between rustc and cargo.  For rustc 1.10.0, they list cargo 0.11, which is what we're building in this request.  Does that help?

jmesmon:
on cargo restrictions: Ok, so we're just in need of a newer rustc to build a newer cargo (I was concerned the issue might have been that a newer cargo was unable to build rustc (well, for our purposes, build `std`)).

srwalter:
Correct.  Life will get better when we upgrade rustc.  We can then build newer cargo that supports more features that will let us remove some of the remaining hacks.

Lemme know if you have any other suggestions (just pushed a couple new commits), otherwise I'll merge in the next day or so.  Cheers!

recipes-devtools/cargo/cargo.inc
```diff
@@ -1,6 +1,6 @@
 INHIBIT_CARGO_DEP = "1"
```
jmesmon: Seems like `cargo_util` ignores this? Is it unneeded? I included it initially because cargo would use a cargo snapshot to build itself, and I believed that I needed to avoid adding `cargo-native` unconditionally to avoid a dependency loop (though I don't think I ever checked omitting it).

If `bitbake` is happy with this current set of changes, you should be able to remove this unused variable.

srwalter: Hmm, that's interesting.  You're correct that this doesn't seem to do anything, but I don't understand how we're not ending up with a circular dependency on cargo-native.  It looks like BASEDEPENDS gets dropped out of DEPENDS on a -native build (which probably means if we build a native package using cargo, it wouldn't get the right dependencies).

I'll play with this a little more.  I'm not sure adding to BASEDEPENDS is a kosher thing to do anyway, as that variable isn't actually documented by bitbake...

srwalter: Went ahead and just removed INHIBIT_CARGO_DEP as you recommended

cardoe:
@srwalter So I'll just say since there's a number of people that can commit now and review things now its probably poor form to merge your own PRs. I actually never got a chance to look at this because it was a long weekend in the US and I was on vacation so I just got back and saw this.

srwalter:
Yes, I was also on vacation, that's why I had time to work on rust :smile:  I waited until I had feedback from another developer; if you'd prefer that I wait for you specifically to give feedback, just say so.

cardoe:
Nope. @jmesmon could have merged it. I was more commenting on the vacation bits as this is why you didn't get others merging it potentially but @jmesmon did comment on it.

srwalter:
Fair enough.  I will try to restrain my exuberance in the future :smile:

PR #72, opened 2016-09-03
compiler-rt: ensure -e is passed to make

tylerwhall:
OE-Core rev aeb653861a0ec39ea7a014c0622980edcbf653fa (between jethro and
krogoth) removed -e from the default flags in EXTRA_OEMAKE. Without
this, the makefile will default to gcc for CC.


jmesmon:
Looks fine to me. Looking at the other projects, it seems like the longer term solution is to fix compiler-rt's makefile to use env vars where needed (via `?=` and variations).

PR #85, opened 2016-09-25
Have rust build its llvm itself

### LINKED ISSUE #80: re-merge rust-llvm-native & rust-native
Now that we build std for target instead of rebuilding the whole compiler, splitting out rust-llvm doesn't save us any build time.

Should we merge them back together?

jmesmon:
Fixes #80 .

This also appears to handle #47 for me (at least with a gcc 5.2 host).


cardoe:
@jmesmon I'll give this a whirl on Monday or Tuesday.

cardoe:
Well I've been able to build this and verify it builds something that works so 👍. But WOW it takes long to build `rust-native` now. I was able to rebuild every other component before this completed. Anyone else want to weigh in on the change? @derekstraka? @srwalter?

srwalter:
Strange that this makes the build take longer.  I would think that we're building the same amount of code either way.  Perhaps this builds LLVM with a lower -j level or something?

The reason I pulled rust-llvm into its own recipe was to reduce the amount of code that had to be built between rust-native and various rust-cross flavors; rust-llvm-native was common to all of them.  However, now that the only real compiler is rust-native, that reasoning no longer applies.

So, given that, I don't have a problem with these changes.  It would be nice to better understand the build time penalty, though :-)

cardoe:
Its similar on Gentoo. Its quicker to build the Gentoo LLVM package and build the Rust package against the system LLVM than building Rust with its shipped LLVM.

jmesmon:
It's possible parallelization isn't being done properly for the llvm sub-build. I'll take a quick look.

I suppose we could consider keeping the rust-llvm separate to add another "cancellation point" of sorts: if part of the rust build fails, we'll still have the rust-llvm build done (potentially), and won't need to rebuild it.

In updating to 1.12.0 ( #89 ), I've noticed that the support that rust has for building llvm isn't that great (not passing flags through), so it might be less work in the long run to take more control of that build process and do it ourselves.

Any thoughts on supporting using a generic llvm for rustc and how that might affect using a seperate package for rust-llvm? (though meta-oe doesn't seem like it's packaging a new enough version at the moment)

cardoe:
So one argument against this would be that LLVM doesn't often change for Rust so we could probably keep the same build of LLVM and get newer versions of Rust built against it.

jmesmon:
I'm planning on taking a look at the llvm support that is in meta-oe already & see if we can use "normal llvm" plus the few backport patches that rust needs instead of building our own. Should also help out other folks who want to use llvm within oe.

srwalter:
FYI, the release notes for rust 1.12 say it should build against LLVM 3.9

jmesmon:
Closing this for now as it seems like going the mainline-llvm route would be more beneficial. If we do ever want to have rustc build llvm though, it's a simple matter of removing the depend and a single configure flag.

On the release notes: It may _build_ with plain 3.9, the problem is what happens afterwards :) . There are some codegen bugs that rust's shipped llvm has backports to fix: https://github.com/rust-lang/rust/issues/36023 (and a few optimization patches). So we'll probably need to include a few patches to apply to the release llvm version.

PR #89, opened 2016-09-30
[WIP] Rust 1.12.0

### LINKED ISSUE #70: Update to rust 1.12
Anyone know when 1.12 is due? Might make sense to plan ahead and target that.

jmesmon:
This is based on my current PR to remove rust-llvm (and have rust build it's llvm) #85 .

Right now, rust-native builds, but libcompiler-rt (for target) fails to build. May simply be some new build system around compiler-rt.

When complete, closes #70


cardoe:
I think you'll have to update the cargo fetcher as well. There's a hardcoded hash in there that changed but in reality we want to use the new support of having a "local" cargo repo. We'll download to the local repo and point cargo to that local repo. That'll be a lot better than trying to emulate the internal directory structure that cargo uses. The support for the "local" cargo repo debuted with Cargo 0.13 which was the companion release with Rust 1.12.0.

cardoe:
We'll also be able to utilize the new `--frozen` flag which debuted with Cargo 0.12 to rather than patching out the network access.

cardoe:
I'm going to close this since #97 supersedes it.

PR #91, opened 2016-09-30
libgit2: add missing dependencies

cardoe:
Without these dependencies cargo is unable to fetch packages from SSH
based repos. This same fix is already upstream in meta-oe master at:
http://cgit.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/libgit2/libgit2_0.24.1.bb


derekstraka:
:+1:

PR #92, opened 2016-10-01
update case of OpenEmbedded to match upstream

cardoe:


cardoe:
If someone agrees with this I'll push the same patch into all of our other branches.

Issue #95, opened 2016-10-26
rust shared source / package split


cardoe:
So I'll be real honest and admit that I don't understand why we do the shared source and split up so much of the configuration. I find the setup here a bit complex and hard to manage. In Gentoo where I maintain the Rust packages as well I've got everything split into a `rustllvm` and `rust`. I extract the tarball twice. I build the target mk files within the sources rather than try to provide an alternative directory. Effectively all I'm patching in Gentoo is the fetching of the stage0 tarball since I have the package manager do that and place it in the right location. What I'm getting at is can we refactor into a similar setup since to me conceptually it will be easier to make version bumps or what am I missing behind some of the separation and other patches.

/cc @jmesmon @derekstraka @srwalter

jmesmon:
IIRC, I originally did the source-split to avoid needing to have 3 separate copies of the source code around (`-native`, `-llvm`, `-cross`). As gcc was doing the same thing, I figured it might be a good idea to do so to reduce size/inodes/path-lookups/uncached-reads/extra-patch-applies. At the time I was also unsure if bitbake would properly avoid downloading the same source tarball 3 times over if something like this was not done.

Note that rebasing the patches is (from my perspective) a fairly easy part of updating meta-rust. In fact, that's most of what the WIP for 1.12 ( #89 ) has in it already (rebased patches). Hold up right now is spending time figuring out how to build new things for target (compiler-rt), and likely figuring out the cargo update. My feeling is the main complexity in updating meta-rust is handling cross compilation while strictly following the specification of the target (flags, features, etc) given by the OE metadata (rust tends to love it's magic configuration).

Also, IIRC, only 1 patch ( https://github.com/meta-rust/meta-rust/blob/master/recipes-devtools/rust/files/rust/0010-mk-allow-changing-the-platform-configuration-source-.patch ) is actually added for handling out-of-tree builds correctly.

One key patch that was (the code around it doesn't get updated much anymore, so it's a bit easier these days) a pain to maintain is the one fixing `libdir` and `bindir` (which I didn't upstream at the time because it would break windows, and it needs support for continuing to use rust's "magic libdir" strategy for default builds). This was absolutely required when we were building a `-cross` compiler, but is somewhat less crucial now as long as everyone uses default paths for everything. I'm not quite sure I want to start making that assumption.

cardoe:
[...]

Issue #81, opened 2016-09-19
Is it possible to build rustc and cargo for target?


kteza1:
I tried using rustup to be able to directly compile on target board and it didn't work. Is it possible to build rustc and cargo using meta-rust?

cardoe:
In theory yes, in practice I'm not sure if anyone has actually tested that.

kteza1:
I'll test that. All I've to do is include `rustc` and `cargo` to image bb file right?

cardoe:
Should be

cardoe:
@kteza1 I'm assuming this has not worked for you? I just did `bitbake rust-llvm` and it failed to build for me.

kteza1:
@cardoe Oops sorry for not getting back. I tried this some time back and it failed I think. I'll report back if I try this again.

jmesmon:
I've tried to do this too with the experiment in merging the rust-llvm build into rust. IIRC, it failed there too due to llvm not recognizing the target it was being built for. It likely needs to be passed an llvm-flavor target instead of a OE target.

cardoe:
So in my branch to build rust-llvm with cmake I've managed to get it compiled for the target. I haven't tried to build rust with it yet. But figured I'd provide an update.

cardoe:
While #105 fixes rust-llvm it does not fix rust which still fails with the following:

```
-isystem/var/build/build/tmp/sysroots/x86_64-linux/usr/include -O2 -pipe -D_GLIBCXX_USE_CXX11_ABI=0  --sysroot=/var/build/build/tmp/sysroots/qemux86-64  -I /var/build/build/tmp/sysroots/x86_64-linux/usr/include -I /var/build/build/tmp/work/core2-64-poky-linux/rust/1.10.0-r0/rustc-1.10.0/src/rustllvm/include /var/build/build/tmp/work/core2-64-poky-linux/rust/1.10.0-r0/rustc-1.10.0/src/rustllvm/RustWrapper.cpp
In file included from /var/build/build/tmp/sysroots/x86_64-linux/usr/include/llvm/ADT/Hashing.h:48:0,
                 from /var/build/build/tmp/sysroots/x86_64-linux/usr/include/llvm/ADT/ArrayRef.h:13,
                 from /var/build/build/tmp/sysroots/x86_64-linux/usr/include/llvm/IR/IRBuilder.h:18,
[...]

PR #97, opened 2016-10-28
Rust 1.12.1

### LINKED ISSUE #70: Update to rust 1.12

cardoe:
Add Rust 1.12.1. Fixes #70

### DIFF
--- a/recipes-devtools/cargo/cargo_0.13.0.bb
+++ b/recipes-devtools/cargo/cargo_0.13.0.bb
@@ -1,0 +1,89 @@
+require cargo-snapshot.inc
+require cargo.inc
+
+SRC_URI += " \
+	git://github.com/rust-lang/cargo.git;protocol=https;name=cargo \
+	git://github.com/rust-lang/rust-installer.git;protocol=https;name=rust-installer;destsuffix=${S}/src/rust-installer \
+	crate://crates.io/advapi32-sys/0.2.0 \
+	crate://crates.io/aho-corasick/0.5.2 \
+	crate://crates.io/bitflags/0.1.1 \
+	crate://crates.io/bitflags/0.7.0 \
+	crate://crates.io/bufstream/0.1.2 \
+	crate://crates.io/cfg-if/0.1.0 \
+	crate://crates.io/cmake/0.1.17 \
+	crate://crates.io/crossbeam/0.2.9 \
+	crate://crates.io/curl-sys/0.2.1 \
+	crate://crates.io/curl/0.3.2 \
+	crate://crates.io/docopt/0.6.82 \
+	crate://crates.io/env_logger/0.3.4 \
+	crate://crates.io/filetime/0.1.10 \
+	crate://crates.io/flate2/0.2.14 \
+	crate://crates.io/fs2/0.2.5 \
+	crate://crates.io/gcc/0.3.32 \
+	crate://crates.io/gdi32-sys/0.2.0 \
+	crate://crates.io/git2-curl/0.5.0 \
+	crate://crates.io/git2/0.4.4 \
+	crate://crates.io/glob/0.2.11 \
+	crate://crates.io/hamcrest/0.1.0 \
+	crate://crates.io/idna/0.1.0 \
+	crate://crates.io/kernel32-sys/0.2.2 \
+	crate://crates.io/lazy_static/0.2.1 \
+	crate://crates.io/libc/0.2.15 \
+	crate://crates.io/libgit2-sys/0.4.5 \
+	crate://crates.io/libressl-pnacl-sys/2.1.6 \
+	crate://crates.io/libssh2-sys/0.1.38 \
+	crate://crates.io/libz-sys/1.0.5 \
+	crate://crates.io/log/0.3.6 \
+	crate://crates.io/matches/0.1.2 \
+	crate://crates.io/memchr/0.1.11 \
+	crate://crates.io/miniz-sys/0.1.7 \
+	crate://crates.io/miow/0.1.3 \
+	crate://crates.io/net2/0.2.26 \
+	crate://crates.io/nom/1.2.4 \
+	crate://crates.io/num-bigint/0.1.33 \
+	crate://crates.io/num-complex/0.1.33 \
+	crate://crates.io/num-integer/0.1.32 \
+	crate://crates.io/num-iter/0.1.32 \
+	crate://crates.io/num-rational/0.1.32 \
+	crate://crates.io/num-traits/0.1.34 \
+	crate://crates.io/num/0.1.34 \
+	crate://crates.io/num_cpus/1.0.0 \
+	crate://crates.io/openssl-sys-extras/0.7.14 \
+	crate://crates.io/openssl-sys/0.7.14 \
+	crate://crates.io/openssl/0.7.14 \
+	crate://crates.io/pkg-config/0.3.8 \
+	crate://crates.io/pnacl-build-helper/1.4.10 \
+	crate://crates.io/rand/0.3.14 \
+	crate://crates.io/regex-syntax/0.3.4 \
+	crate://crates.io/regex/0.1.73 \
+	crate://crates.io/rustc-serialize/0.3.19 \
+	crate://crates.io/semver/0.2.3 \
+	crate://crates.io/strsim/0.3.0 \
+	crate://crates.io/tar/0.4.8 \
+	crate://crates.io/tempdir/0.3.5 \
+	crate://crates.io/term/0.4.4 \
+	crate://crates.io/thread-id/2.0.0 \
+	crate://crates.io/thread_local/0.2.6 \
+	crate://crates.io/toml/0.2.0 \
+	crate://crates.io/unicode-bidi/0.2.3 \
+	crate://crates.io/unicode-normalization/0.1.2 \
+	crate://crates.io/url/1.2.0 \
+	crate://crates.io/user32-sys/0.2.0 \
+	crate://crates.io/utf8-ranges/0.1.3 \
+	crate://crates.io/winapi-build/0.1.1 \
+	crate://crates.io/winapi/0.2.8 \
+	crate://crates.io/ws2_32-sys/0.2.1 \
+"
+# Compatible with Rust 1.12.1
+# https://static.rust-lang.org/dist/channel-rust-1.12.1.toml
+SRCREV_cargo = "109cb7c33d426044d141457049bd0fffaca1327c"
+
+SRCREV_rust-installer = "c37d3747da75c280237dc2d6b925078e69555499"
+
+S = "${WORKDIR}/git"
+
+LIC_FILES_CHKSUM ="\
+	file://LICENSE-MIT;md5=362255802eb5aa87810d12ddf3cfedb4 \
+	file://LICENSE-APACHE;md5=1836efb2eb779966696f473ee8540542 \
+	file://LICENSE-THIRD-PARTY;md5=892ea68b169e69cfe75097fc38a15b56 \
+"
--- a/recipes-devtools/rust/compiler-rt_1.12.1.bb
+++ b/recipes-devtools/rust/compiler-rt_1.12.1.bb
@@ -1,0 +1,24 @@
+SUMMARY = "Rust compiler run-time"
+HOMEPAGE = "http://www.rust-lang.org"
+SECTION = "devel"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://src/compiler-rt/LICENSE.TXT;md5=bf24bca27049b52e9738451aa55771d4"
+
+require rust.inc
+require rust-source-${PV}.inc
+
+DEPENDS += "rust-llvm-native (=${PV})"
+
+S = "${WORKDIR}/rustc-${PV}"
+
+DISABLE_STATIC = ""
+INHIBIT_DEFAULT_RUST_DEPS = "1"
+
+do_compile () {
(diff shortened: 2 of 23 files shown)


recipes-devtools/rust/rust-version.inc
```diff
@@ -1,6 +1,6 @@
 # Note: if you adjust this, you'll also need to change the hashes in
 # rust-source.bb
 SOURCE_NAME = "rust"
-PV = "1.10.0"
+PV = "1.12.0"
```
posborne: Please note that 1.12.1 has been released and is probably the one you want to be targeting.

cardoe: Sure. This branch is literally me taking @jmesmon's work in #89 and pulling it into my own repo and rebasing it on top of master + the CI work I've been doing.

cardoe: I should also note I haven't made any changes. I just wanted to test another branch tickling GitHub + Jenkins in the right way.

posborne: > I just wanted to test another branch tickling GitHub + Jenkins in the right way.

No problem, happy to see all of the recent work that has been done on the CI front and continuing to work with upstream to reduce patches that are required.  As a stopgap for supporting various stable versions of rust with a more regular cadence @nastevens and I are currently maintaining the [meta-rust-bin](https://github.com/rust-embedded/meta-rust-bin) layer which provides similar functionality but using snapshot releases for host/target from the official mirrors.

Ultimately, we believe the meta-rust layer takes the right approach but found that in the interim going another route met our immediate needs best since our target happens to be one of the ones for which rust provides host/target executables which are "good enough" for our (and perhaps most) use cases.

cardoe:
My use case currently works fine with the binary releases so I was thinking of adding the binary releases to the layer and giving people the option of using those. 

My ultimate goal is to make bumps easier as well. Which was the point of me asking questions in #95. I've iterated the Gentoo process to the point of 1 small patch and it has applied the last 3 releases cleanly. So it's as simple as renaming the recipe file to bump the version. That's where we need to be. 

--
Doug

posborne:
> My use case currently works fine with the binary releases so I was thinking of adding the binary releases to the layer and giving people the option of using those.

Sounds good to me.  I *think* that wouldn't be too hard via `PREFERRED_PROVIDER` and `PREFERRED_VERSION` for rust/cargo.  This might work out well as cargo/rust snapshot releases are required to build rustc/std from source anyway.

> So it's as simple as renaming the recipe file to bump the version. That's where we need to be.

Agreed on needing to make it very simple to bump versions.  I talked with @alexcrichton briefly at Rust Belt Rust and we generally agreed it is in everyone's interest to reduce the number of patches required in layers like meta-rust and other similar build systems.  Seems like we are doing a good job of working toward that but with the some uncertainties on how to get rid of a few patches due to uncertainty about whether/when things are moving to rustbuild uniformly.

For the bin packages, we got things down to the point where we just run [this script](https://github.com/rust-embedded/meta-rust-bin/blob/master/build-new-version.sh) to support a new version of rust/cargo in meta-rust-bin.

cardoe:
@jmesmon So this branch is now at a compiler-rt failure. I think you had mentioned seeing that as well. Did you come up with a solution?

jmesmon:
So, I had looked at how compiler-rt was being built in 1.12, and IIRC, rust builds it either via some custom stuff in `mk/rt.mk` which selects the source files based on guessing what the triple wants, or via a rust build script that does much the same thing.

In 1.13 (or 1.14) I believe the build script was moved around so compiler-rt can be built like a normal crate, but in 1.12 it's unfortunately integrated into `src/bootstrap/native.rs`, and isn't clear that it is usable outside the rustbuild stuff (which we'd likely need to do as we're not building the entire compiler).

Easiest solution is likely to patch up `mk/rt.mk` and call it to build compiler-rt. Harder one would be back-porting the changes from 1.13 or 1.14 to allow compiler-rt to build as a crate.

cardoe:
I'll be gone for the next week but when I return I'll look at 1.13.

--
Doug

jmesmon:
Just a note, 1.13 has a bug on arm hf systems that we'll need to backport the fix for (as it's rather likely folks want arm hf out of meta-rust): https://github.com/rust-lang/rust/issues/37630

cardoe:
I believe this next build will have this working. @meta-rust/maintainers would anyone like to review this?

recipes-devtools/cargo/cargo_0.13.0.bb
```diff
@@ -0,0 +1,125 @@
...
+	file://LICENSE-MIT;md5=362255802eb5aa87810d12ddf3cfedb4 \
+	file://LICENSE-APACHE;md5=1836efb2eb779966696f473ee8540542 \
+	file://LICENSE-THIRD-PARTY;md5=892ea68b169e69cfe75097fc38a15b56 \
+"
+
+## curl-rust
+SRC_URI += "\
+       git://github.com/carllerche/curl-rust.git;protocol=https;destsuffix=curl-rust;name=curl-rust \
+       "
+#       file://curl-rust/0001-curl-sys-avoid-explicitly-linking-in-openssl.patch;patchdir=../curl-rust \
```
jmesmon: It looks like this and most (all?) other patches to cargo deps are commented out, can the overrides be removed entirely? (pulling from git repos for some deps was only added so we could apply patches, starting way back in the old-style cargo builds).

On a similar note: is it possible to apply patches to sources from `crate://`?

recipes-devtools/rust/compiler-rt_1.12.1.bb
```diff
@@ -0,0 +1,60 @@
...
+S = "${WORKDIR}/rustc-${PV}"
+
+DISABLE_STATIC = ""
+INHIBIT_DEFAULT_RUST_DEPS = "1"
+
+do_configure () {
+    # rust.inc generates these into ${WORKDIR}/mk-cfg/ but
+    # save ourselves time and patching by moving them back
+    # to the normal spot
+    mv "${WORKDIR}"/mk-cfg/*.mk "${S}/mk/cfg/" || bbfatal "couldn't move platform cfg"
```
jmesmon: This really seems like an indication that rust-source-${PV}.inc should include a SRC_URI with the rust source and all our patches for that version. Would let us just use the same configure args as the initial build.

recipes-devtools/cargo/cargo_0.13.0.bb
cardoe: ah yes. you're right. I meant to strip this all out. I wasn't actually sure why we prevented linking to OpenSSL as it works. Technically if we ever wanted to build cargo for the target like #81 requests then we wouldn't have a very featureful cargo.

Right now I don't have a way to patch `crate://` baked in. Its feasible since `crate://` just un-tars the the crate now.

recipes-devtools/rust/compiler-rt_1.12.1.bb
jmesmon: Or if not SRC_URI, some variable used by other recipes to get the same source.

cardoe: Flip side would be to remove that patch and just put them in ${S}/mk/cfg/ and no patching needed for compiler-rt. It'll be unnecessary come 1.13.0.

recipes-devtools/rust/rust_1.12.1.bb
```diff
@@ -0,0 +1,26 @@
+inherit rust-installer
+require rust.inc
+require rust-source-${PV}.inc
+
+DEPENDS += "cmake-native"
```
jmesmon: Don't we only need cmake for rust-llvm, not rust itself?

recipes-devtools/rust/compiler-rt_1.12.1.bb
jmesmon: Still, imagine we need to carry a patch for libstd or compiler-rt: we'd want it to be applied to both native (build by rust-native) and target (built by libstd-rs and compiler-rt).

Having a common SRC_URI or other variable makes it much harder to get that wrong.

recipes-devtools/rust/rust_1.12.1.bb
cardoe: No. The configure script checks for cmake and doesn't allow the configure script to succeed without it. Approximately line 850.

```
# For building LLVM
probe_need CFG_CMAKE cmake
```

recipes-devtools/rust/compiler-rt_1.12.1.bb
```diff
@@ -0,0 +1,60 @@
...
+        ${EXTRA_OECONF}
+}
+
+do_compile () {
+	oe_runmake ${TARGET_SYS}/rt/libcompiler-rt.a
+}
+
+do_install () {
+	mkdir -p ${D}${libdir}
+	cp ${TARGET_SYS}/rt/libcompiler-rt.a ${D}${libdir}/libcompiler-rt.a
```
jmesmon: Sure, that's fine then. I've probably asked about this previously too :)

recipes-devtools/rust/rust_1.12.1.bb
jmesmon: Ah, a bug in the configure script then? (shouldn't check if it isn't going to be building llvm, and in our case it isn't).

Sounds fine, might be good to note that it is only included to workaround that.

recipes-devtools/rust/compiler-rt_1.12.1.bb
cardoe: You did at https://github.com/meta-rust/meta-rust/pull/51#issuecomment-243815407

recipes-devtools/rust/rust_1.12.1.bb
cardoe: I can't find the issue for it but I've raised it with upstream and the patch is stuck in rollup queues. It didn't seem harmful to include it since rust-llvm depends on it so we've already built it.

recipes-devtools/rust/compiler-rt_1.12.1.bb
cardoe: Oh I'm not disagreeing about the common SRC_URI. I'll update it.

jmesmon: It could also be a good idea to push the configure arguments into a common location (presuming they're the same).

But that's less important if we're switching to cargo based builds here in 1.13.0,

recipes-devtools/rust/rust_1.12.1.bb
jmesmon: Yep, sounds fine.

recipes-devtools/cargo/cargo_0.13.0.bb
```diff
@@ -0,0 +1,125 @@
...
+	crate://crates.io/advapi32-sys/0.2.0 \
+	crate://crates.io/aho-corasick/0.5.2 \
+	crate://crates.io/bitflags/0.1.1 \
+	crate://crates.io/bitflags/0.7.0 \
+	crate://crates.io/bufstream/0.1.2 \
+    crate://crates.io/cfg-if/0.1.0 \
+	crate://crates.io/cmake/0.1.17 \
+	crate://crates.io/crossbeam/0.2.9 \
+	crate://crates.io/docopt/0.6.82 \
+    crate://crates.io/env_logger/0.3.4 \
```
jcreekmore: Spacing issue

recipes-devtools/cargo/cargo_0.13.0.bb
```diff
@@ -0,0 +1,125 @@
...
+	crate://crates.io/filetime/0.1.10 \
+	crate://crates.io/flate2/0.2.14 \
+	crate://crates.io/fs2/0.2.5 \
+	crate://crates.io/gcc/0.3.32 \
+	crate://crates.io/gdi32-sys/0.2.0 \
+	crate://crates.io/glob/0.2.11 \
+	crate://crates.io/hamcrest/0.1.0 \
+	crate://crates.io/idna/0.1.0 \
+	crate://crates.io/kernel32-sys/0.2.2 \
+    crate://crates.io/lazy_static/0.2.1 \
```
jcreekmore: Actually, lots of spacing issues... check the rest of this list, because there are more.

cardoe:
So it looks like it only failed because crates.io was returning 500's when fetching crates

recipes-devtools/cargo/cargo_0.13.0.bb
cardoe: Fixed. Thanks.

recipes-devtools/cargo/cargo_0.13.0.bb
cardoe: Rewrote this to not use the SRC_URI's for the git repos. The fetching of the SRC_URIs like this would be the proper way to apply patches to crates IMHO. But since we don't need to patch them anymore I've removed it.

cardoe:
This merges cleanly. Its just missing the merge commit from #123. It contains all the commits of #123. @jmesmon Do you want to peek at these changes and make sure they address all your comments?

We don't have Jenkins setup the best because its auto rebasing branches and retesting them but that's not being reflected in GitHub. So the original push I did is actually https://ci.asterius.io/job/meta-rust/job/PR-97/45/ which succeeded but https://ci.asterius.io/job/meta-rust/job/PR-97/46/ is rebased on master as it currently is. So should both succeed they can be merged (I would do a rebase and merge from the drop down personally).

PR #101, opened 2016-11-01
Drop shared source directory

cardoe:
This drops the shared source directory. It also adds the version number to the bitbake recipe file name so in the future we can have multiple versions through `PREFERRED_VERSION`. ref #95


cardoe:
Running one more build to have 3 builds before this branch and 3 builds of this branch to average out the times to show the difference. The metric I'm using is `time bitbake rust-native` and in between runs I've done `bitbake -c cleansstate rust-llvm-native rust-native`.

cardoe:
Average time for dropped share source: 44:14
Average time for original: 43:58

The values are in minutes:seconds.

jmesmon:
16 seconds is fairly small. I presume this timing was done on the jenkins infra?

If so, due to that system using a ram disk for the build directory, it's likely avoiding a bit of slowdown that others (without a 64GiB ram disk) will see.

That said, I'm not sure it's worth spending time doing more timing & sizing of builds. I'm not against merging this if folks want it.

Tangental Jenkins question: I notice that SSTATE_MIRROR and SOURCE_MIRROR_URL are set, but it's unclear if those locations are actually being populated & used by builds. Are they? (shouldn't have any effect on this timing, just curious).

cardoe:
It was done on my work laptop. core i5 with 16gb RAM. I'm not using tmpfs or anything special so I bet it will be faster on Jenkins.

derekstraka:
I originally populated downloads and SSTATE cache by hand from the builds I was running from master (That's why master PRs get a substantial decrease in build time).  My intention is to have the nightly builds clean out and then populate the caches for each major branch.  That way we end up getting a clean build from scratch every night.  Basically, PRs would be read only to the cache to avoid any squirrely behavior

derekstraka:
I'm good with this to be merged

PR #103, opened 2016-11-02
Small cleanups

cardoe:
Some additional cleanups I've queued after #101 lands.

- combine some `rust-llvm` bits that never made it into master and have been in fido and the starlab-io fork.
- fix the fetch script that Jenkins uses if a user decides to use it for testing to default to a sensible value
- change the default target machine to x86_64 from i586.

This is based on #101.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 *.pyc
+build/
+poky/
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -35,7 +35,7 @@
 #MACHINE ?= "edgerouter"
 #
 # This sets the default machine to be qemux86 if no other machine is selected:
-MACHINE ??= "qemux86"
+MACHINE ??= "qemux86-64"
 
 #
 # Where to place downloads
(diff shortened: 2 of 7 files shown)


srwalter:
What does it buy us to remove the old style recipes?

cardoe:
Well they're dependencies and with the new way to build things how can they be consumed?

cardoe:
I can drop their removal from this branch if you want.

srwalter:
I would prefer that for now.  I have my own recipes not in meta-rust (but used in production) that depend on some of them.

cardoe:
@srwalter Are you depending on master in production, not pinned to a specific revision?

cardoe:
Anyone have any more feedback? Otherwise I think this is good to go.

tylerwhall:
:+1:

srwalter:
No, it won't break me immediately, but I do update frequently

srwalter:
:+1:

cardoe:
@srwalter Well I'd appreciate if you have some time to come up with a test case that we can integrate into Jenkins to ensure we don't break your workflow.

srwalter:
That's a good idea, I'll see what I can come up with

PR #104, opened 2016-11-03
rust: fix LLVM data layouts for Rust 1.10

cardoe:
The data layout changed for Rust 1.10. I noticed this because I got a
crash when using rust-native to build rust. The assertion that hit was
saying data layout mismatch and caused me to check the values in the
Rust sources.

### DIFF
--- a/recipes-devtools/rust/rust.inc
+++ b/recipes-devtools/rust/rust.inc
@@ -126,35 +126,35 @@ def llvm_features_from_cc_arch(d):
     return ','.join(f)
 
 ## arm-unknown-linux-gnueabihf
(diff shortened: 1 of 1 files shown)


cardoe:
For reference this is where I got the values from:

- https://github.com/rust-lang/rust/blob/1.10.0/src/librustc_back/target/aarch64_unknown_linux_gnu.rs#L17
- https://github.com/rust-lang/rust/blob/1.10.0/src/librustc_back/target/arm_unknown_linux_gnueabihf.rs#L20
- https://github.com/rust-lang/rust/blob/1.10.0/src/librustc_back/target/i686_unknown_linux_gnu.rs#L23
- https://github.com/rust-lang/rust/blob/1.10.0/src/librustc_back/target/x86_64_unknown_linux_gnu.rs#L23

tylerwhall:
:+1:

srwalter:
:+1:

cardoe:
The 2nd commit needs to be backported to krogoth and jethro. The first commit is specific to Rust 1.10 as far as I can tell from commit history.

PR #105, opened 2016-11-04
convert rust-llvm to be built with cmake

cardoe:
Starting with Rust 1.12.0, the Rust build has switched to using LLVM 3.9. LLVM 3.9 and newer will only build with cmake and have removed their autotools build system. In anticipation of that and allowing us to roll forward more easily I've converted the build of rust-llvm to using cmake. This fixes the build issue if rust-llvm for targets that #81 mentions but I'm not yet sure if it fixes the issue entirely. I purposefully did this for Rust 1.10.0 so that we can confirm that there is no code generation differences when changing LLVM from autotools to cmake. One change I have noticed is that rust-llvm is now built with assertions but I'm leaving those on because they caught actual bugs in our build that existed in our Rust 1.10.0 build.


srwalter:
:+1:

PR #106, opened 2016-11-08
More updates

cardoe:
- Refactor bits to eventually allow for multiple versions of Rust in the tree
- Drop PACKAGECONFIG for local-rust and add possible support for multiple build architectures.

PR #107, opened 2016-11-09
rust: drop bindir/libdir patch

cardoe:
I'm honestly not sure what this patch does since everything appears to
build the same with and without it. I did a `tree` of the sysroots and
images and I don't see any file differences. Something like this used to
be necessary in Gentoo but I dropped it a few releases back.


cardoe:
This is just one patch on top of #106.

jmesmon:
It probably will be fine to drop as long as the libdir & bindir in bitbake.conf stay at their current values. Might be an edge case around 64bit stuff (as without this patch rust tries to select between `lib` & `lib64` with it's own internal logic, instead of using the specified on from bitbake), but I'm not sure (and the only special case wrt that in bitbake.conf is powerpc64, which I've never tested building meta-rust for).

cardoe:
> As a note, original need was because cross.bbclass changes bindir and libdir so cross compilers end up seperate from host compilers (ex: /usr/bin becomes /usr/${TARGET_SYS}/bin and similar). As the cross compiler is gone, we're no longer restricted by cross.bbclass.

Ah. That makes perfect sense. But I hope you agree that less is more now that we've done away with the cross compiler?

cardoe:
@jmesmon, @posborne: Feel free to merge if people are happy with this change.

PR #108, opened 2016-11-09
drop rust-installer.bbclass

cardoe:
Rework cargo to avoid using rust-installer and integrate the rust-installer cleanups into the rust do_install() to avoid having another bbclass


jmesmon:
Isn't the `do_install_append()` functional even if EXPORT_FUNCTIONS isn't? Perhaps this should be merged into one of the rust includes?

recipes-devtools/rust/rust.inc
```diff
@@ -452,6 +452,12 @@ rust_do_install () {
     # Remove any files directly installed into libdir to avoid
     # conflicts between cross and native
     rm -f ${D}${libdir}/lib*.so
+
+    # cleanup after rust-installer since we don't need these bits
+	rm ${D}/${libdir}/rustlib/install.log
```
jmesmon: weird spacing (which I'm sure is all over these files, sorry :)

cardoe:
I've pulled the first commit into another branch and I'll roll your suggested changes which were to the 2nd patch into another branch.

PR #114, opened 2016-11-21
Improve Rust triple generation

cardoe:
Improve how we generate the native Rust triples from `TARGET_SYS`, `BUILD_SYS`, and `HOST_SYS` to hopefully more accurately generate these. This should solve issues with i586/i686 and musl based systems. Hopefully this future proofs the behavior for further support landing in upstream.

### DIFF
--- a/classes/rust-triples.bbclass
+++ b/classes/rust-triples.bbclass
@@ -9,10 +9,9 @@ def rust_base_triple(d, thing):
     '''
 
     arch = d.getVar('{}_ARCH'.format(thing), True)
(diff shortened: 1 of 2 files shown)


cardoe:
@meta-rust/maintainers I believe this should produce proper triples for the built-in Rust triples and the future triples being added. I've actually got a further branch that switches away from using custom triples to mapping to the built-in triples while still allowing custom triples that are provided via a bbappend instead of generating custom triples everytime. That relies on this branch. I figured that was a bit more controversial so we'll discuss that separately.

PR #116, opened 2016-11-23
refactor linker flags out of custom targets

cardoe:
Refactors the Yocto linker flags out of the custom targets and into the rustflags option of the cargo config. This is part of my series to eventually allow us to use the native targets and slim down the delta between upstream Rust/Cargo and what's in meta-rust.

### DIFF
--- a/classes/cargo.bbclass
+++ b/classes/cargo.bbclass
@@ -1,6 +1,7 @@
 inherit rust-vars
 # add crate fetch support
 inherit crate-fetch
+inherit rust-triples
 
 # the binary we will use
 CARGO = "cargo"
@@ -47,6 +48,15 @@ local-registry = "${WORKDIR}/cargo_registry"
 replace-with = "local"
 registry = "https://github.com/rust-lang/crates.io-index"
 EOF
+
+    # We need to use the real Yocto linker and get the linker
+    # flags to it. Yocto has the concept of BUILD and TARGET
+    # and uses HOST to be the currently selected one. However
+    # LDFLAGS and TOOLCHAIN_OPTIONS are not prefixed with HOST
+    echo "[build]" >> ${CARGO_HOME}/config
+    echo "rustflags = ['-C', 'link-args=${LDFLAGS}${HOST_CC_ARCH}${TOOLCHAIN_OPTIONS}']" >> ${CARGO_HOME}/config
+    echo "[target.${RUST_HOST_SYS}]" >> ${CARGO_HOME}/config
+    echo "linker = '${HOST_PREFIX}gcc'" >> ${CARGO_HOME}/config
 }
 
 # All the rust & cargo ecosystem assume that CC, LD, etc are a path to a single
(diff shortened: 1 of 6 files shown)


cardoe:
@meta-rust/maintainers I'm still in the process of testing this but it should be good enough for a review.

cardoe:
Well locally backing up a few commits in poky results in a qemuarm and qemux86-64 that produce images that work.

classes/cargo.bbclass
```diff
@@ -59,14 +75,13 @@ export RUST_CFLAGS = "${HOST_CC_ARCH}${TOOLCHAIN_OPTIONS} ${CFLAGS}"
 export RUST_BUILD_CC = "${CCACHE}${BUILD_PREFIX}gcc"
 export RUST_BUILD_CFLAGS = "${BUILD_CC_ARCH} ${BUILD_CFLAGS}"
 
-RUSTFLAGS ??= ""
-export CARGO_BUILD_FLAGS = "-v --target ${HOST_SYS} --release"
+CARGO_BUILD_FLAGS = "-v --target ${HOST_SYS} --release"
```
posborne: This could be added in a separate set of changes, but I have needed to modify the cargo build flags for some projects in the past and having something like `EXTRA_CARGO_FLAGS` is useful.  Some of the things we do here would seem useful to mimic (meta-rust-bin also deals with debug/release builds): https://github.com/rust-embedded/meta-rust-bin/blob/master/classes/cargo.bbclass

cardoe: Yeah that's a good idea. Feel free to make a follow on branch or I can queue something up in a future branch.

recipes-devtools/cargo/cargo.inc
```diff
@@ -67,6 +67,10 @@ SRC_URI = "\
 
 B = "${S}"
 
+# Can't use the default from rust-vars as it cause a compiler crash
+# the crate_hash option will cause the compiler to crash
```
srwalter: That's news to me.  How are you ensuring that you're generating usable sstate entries without the crate_hash option?

cardoe:
> Did you test these changes with any recipes that don't use cargo to build?

Please submit a test that does not use cargo to build.

> Your commit message omits the most important part: "why". I can easily see what you're doing without the commit message.

The goal is to cut our magic down significantly so that bumps are trivial. You are right its not mentioned on the commit message but instead of the merge commit.

cardoe:
@srwalter I've also re-imagined these changes as cc and ld wrappers. That might give you more joy with non-cargo builds. Some of the issues with this repo is that a lot of the "why" that you mentioned is missing and along with "how".

jmesmon:
It isn't clear to me that this is really less magic. target specs still seem like they're geared towards exactly what we want to accomplish here: custom flags (and other items) for the linker.

cardoe:
We can revert this. Honestly I'd like to get away from the custom targets for everything. Twice the custom targets in the tree have had data layouts for LLVM that were incorrect and lead to illegal instructions on targets that I've worked with.

recipes-devtools/cargo/cargo.inc
cardoe: See #126.

PR #117, opened 2016-11-25
getVar will expand by default, so the 'True' option is no longer required

derekstraka:


cardoe:
What version of Yocto does this begin the default?

--
Doug

derekstraka:
It is only in [master](https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=2631c375b8a025ee2e4329f6d93900bb7b5dadf7)

derekstraka:
This needs to be merged after the morty branch is created

derekstraka:
The build will be broken until we switch back to master as the default

PR #118, opened 2016-11-28
Continued build updates

### LINKED ISSUE #76: Packages fail to build with 'ccache'
Build fails when 'ccache' is used to build packages; e.g. 'rust-native':

| error: could not exec the linker `ccache gcc`: No such file or directory (os error 2)

or cargo:

| running: "ccache gcc" "-O3" "-ffunction-sections" ...
| failed to execute command: No such file or directory (os error 2)
| Is `ccache gcc` not installed?

cardoe:
Switched to wrapper scripts over using rustflags from #116. This should hopefully address some of the feedback in #116 after it was merged. These changes additionally fix #76.

### DIFF
--- a/classes/cargo.bbclass
+++ b/classes/cargo.bbclass
@@ -62,13 +62,6 @@ EOF
 	echo "linker = '${RUST_TARGET_CCLD}'" >> ${CARGO_HOME}/config
 }
 
-# All the rust & cargo ecosystem assume that CC, LD, etc are a path to a single
-# command. Fixup the ones we give it so that is the case.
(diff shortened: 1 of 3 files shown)


cardoe:
Unfortunately I found an old comment that's wrong in cargo.bbclass with these changes so I rebased that in now.

cardoe:
ok GitHub is acting up because this branch has 7 commits. Its dropping the first one.

cardoe:
I'm at a loss what is going on because I didn't close this branch.

PR #119, opened 2016-11-28
Continued build updates

### LINKED ISSUE #76: Packages fail to build with 'ccache'

cardoe:
Switched to wrapper scripts over using rustflags from #116. This should hopefully address some of the feedback in #116 after it was merged. These changes additionally fix #76.

### DIFF
--- a/classes/rust-triples.bbclass
+++ b/classes/rust-triples.bbclass
@@ -56,3 +56,51 @@ def rust_base_triple(d, thing):
 RUST_BUILD_SYS = "${@rust_base_triple(d, 'BUILD')}"
 RUST_HOST_SYS = "${@rust_base_triple(d, 'HOST')}"
 RUST_TARGET_SYS = "${@rust_base_triple(d, 'TARGET')}"
+
+# wrappers to get around the fact that Rust needs a single
+# binary but Yocto's compiler and linker commands have
+# arguments. Technically the archiver is always one command but
+# this is necessary for builds that determine the prefix and then
+# use those commands based on the prefix.
+WRAPPER_DIR = "${WORKDIR}/wrapper"
+RUST_BUILD_CC = "${WRAPPER_DIR}/build-rust-cc"
+RUST_BUILD_CCLD = "${WRAPPER_DIR}/build-rust-ccld"
+RUST_BUILD_AR = "${WRAPPER_DIR}/build-rust-ar"
+RUST_TARGET_CC = "${WRAPPER_DIR}/target-rust-cc"
+RUST_TARGET_CCLD = "${WRAPPER_DIR}/target-rust-ccld"
+RUST_TARGET_AR = "${WRAPPER_DIR}/target-rust-ar"
+
+create_wrapper () {
+	file="$1"
+	shift
+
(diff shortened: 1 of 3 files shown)


classes/rust-triples.bbclass
```diff
@@ -56,3 +56,68 @@ def rust_base_triple(d, thing):
...
+# compiler is used by gcc-rs
+# linker is used by rustc/cargo
+# archiver is used by the build of libstd-rs
+do_rust_create_wrappers () {
+	mkdir -p "${WRAPPER_DIR}"
+
+	# Yocto Build / Rust Host compiler
+	cat <<- EOF > "${RUST_BUILD_CC}"
+	#!/bin/sh
+	${BUILD_CC} \$@
```
srwalter: $@ needs to be in quotes itself in order to preserve spaces in its arguments.  Otherwise every word will get broken up into its own argument.

classes/rust-triples.bbclass
```diff
@@ -56,3 +56,68 @@ def rust_base_triple(d, thing):
...
+	EOF
+	chmod +x "${RUST_BUILD_CC}"
+
+	# Yocto Build / Rust Host linker
+	cat <<- EOF > "${RUST_BUILD_CCLD}"
+	#!/bin/sh
+	${BUILD_CCLD} ${BUILD_LDFLAGS} \$@
```
srwalter: This is fairly redundant, could make a helper function.  make_wrapper_script ${RUST_BUILD_CCLD} "${BUILD_CCLD ${BUILD_LDFLAGS}", etc

srwalter:
Looks good overall.  At the very least I'd like the quoting in the shell wrappers fixed.

classes/rust-triples.bbclass
cardoe: I assume that was `"${BUILD_CCLD}" "${BUILD_LDFLAGS}"`?

cardoe:
@srwalter Rebased the change in. Let me know if you think quoting should change.

jmesmon:
Using wrappers seems like a sane way to generally solve our issue with gcc-rs not splitting `CC`, `AR`, etc when they contain multiple args.

I'm not entirely sure what the expectations in yocto around flags having spaces is, but I'm not sure anyone will have tried having flags that contain characters to split on. I suppose `cargo` might occasionally pass some, so it would be good to quote if possible (which seems to be happening).

I do wonder how taking this approach with building -native packages would affect the `-nativesdk` and running rust on target, but it as we don't really support those yet it's probably fine to look at that later.

cardoe:
@srwalter Since @jmesmon gave this the thumbs up, if you're happy with the changes please feel free to merge it in.

cardoe:
Well if its ok, can we merge this in?

cardoe:
Rebased

PR #120, opened 2016-11-29
Jenkinsfile: rename stages for easier tracking

cardoe:
Rename the stages to ensure that the machine name is in there so its
easier to track what machine is currently being built.


cardoe:
This one really shouldn't be controversial and I'll gladly merge it myself in the morning.

PR #121, opened 2016-11-30
setup-env: only allow slaves with large amounts of RAM to mount tmpfs

derekstraka:


derekstraka:
I forgot about `mountpoint`; I'll use that instead

cardoe:
TIL about `mountpoint`, thanks @jmesmon!

PR #123, opened 2016-12-02
Drop mk cfg directory change patch

cardoe:
Drops the ${S}/mk/cfg -> ${WORKDIR}/mk-cfg/ patching along with some other outdated bits.


cardoe:
@jmesmon We spoke about dropping this on IRC and it would address one of your review comments on #97. But CC @meta-rust/maintainers for more feedback.

jmesmon:
My only concern is about when (what tasks in yocto) are really supposed to modify or add files to `${S}`, but I'm not aware of a concrete rule in that area.

derekstraka:
There isn't an existing rule about modifying `${S}`.  There are a bunch of recipes out there that modify the sources with a configure or compile prepend depending on what they need to update (configure inputs vs make files).  You are pretty much on your own not to blow yourself up.  Since it is in a common class, we should just document it either inline or as part of the how to

cardoe:
Well @jmesmon if you're ok with @derekstraka's response and ok with this, please click merge? And I'll get #97 rebased and ready to go.

Issue #126, opened 2016-12-05
crasher when using -C crate_hash


cardoe:
## Version(s) of meta-rust

- master / Rust 1.12.1
- krogoth-1.12.1 / Rust 1.12.1

## Version(s) of poky and/or oe-core

- krogoth
- morty

## Steps to reproduce

Hard to reproduce but I've made it happen with two packages. At one point it was happening in this repo with which prompted me to remove `-C crate_hash` from the build of cargo itself.

cardoe:
The crash:

```
| error: internal compiler error: /var/build/build/tmp/work/x86_64-linux/rust-native/1.12.1-r0/rustc-1.12.1/src/librustc_metadata/decoder.rs:74: lookup_item: id not found: DefIndex(101) in crate "uuid" with number 23
|
| note: the compiler unexpectedly panicked. this is a bug.
|
| note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports
|
| note: run with `RUST_BACKTRACE=1` for a backtrace
|
| thread 'rustc' panicked at 'Box<Any>', /var/build/build/tmp/work/x86_64-linux/rust-native/1.12.1-r0/rustc-1.12.1/src/librustc_errors/lib.rs:634
| stack backtrace:
|    1:     0x7fd5edbbff69 - <unknown>
|    2:     0x7fd5edbd07ac - <unknown>
|    3:     0x7fd5edbceba4 - <unknown>
|    4:     0x7fd5edbcf308 - std::panicking::rust_panic_with_hook::h8d486474663979b9
|    5:     0x7fd5e6d337f7 - <unknown>
|    6:     0x7fd5e6d44325 - rustc_errors::Handler::bug::h126e43857fe80736
|    7:     0x7fd5eae30194 - <unknown>
|    8:     0x7fd5ead707d5 - <unknown>
|    9:     0x7fd5ead70612 - rustc::session::bug_fmt::hc0d300eca0d34bd4
|   10:     0x7fd5eb496aa2 - rustc_metadata::decoder::get_adt_def::h2b5e55915402268b
|   11:     0x7fd5eb4b53b4 - rustc_metadata::csearch::_<impl rustc..middle..cstore..CrateStore<'tcx> for rustc_metadata..cstore..CStore>::adt_def::h8d598db111499922
[...]

PR #127, opened 2016-12-06
workaround bug in the -C crate_hash option

cardoe:
The -C crate_hash option does not work if multiple versions of the same
crate are being linked into a binary. This can happen if different
dependencies depend on different versions of the same crate. The -C
crate_hash argument causes all versions of a given crate to have the
same crate hash which in turn causes the compiler to select the wrong
one internally and crash. This works around issue #126 for Cargo builds
but does not fix the underlying cause.

### DIFF
--- a/classes/rust-bin.bbclass
+++ b/classes/rust-bin.bbclass
@@ -3,6 +3,7 @@ inherit rust
 DEPENDS_append = " patchelf-native"
 RDEPENDS_${PN} += "${RUSTLIB_DEP}"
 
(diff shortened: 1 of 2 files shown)


cardoe:
ref #33

cardoe:
Somehow Jenkins and GitHub weren't able to communicate. The build passed successfully but Jenkins wasn't able to update the status on GitHub. Given the two approvals and the fact that this solves an immediate crasher for me I'm going to merge this.

Issue #125, opened 2016-12-05
Create a release branch for morty


derekstraka:
A morty stable branch should be created now that rust 1.12.1 is merged into master

cardoe:
I'd like to do this after #124 lands.

cardoe:
Branched.

PR #128, opened 2016-12-06
drop Rust 1.10.0

cardoe:
This will allow us to do some cleanups with the cargo builder


cardoe:
This depends on #125 and #117.

PR #129, opened 2016-12-08
Merge branch 'morty' into krogoth

### LINKED ISSUE #87: Remove internet access in do_compile for rust and cargo
Doing internet access in do_compile is not a good play with yocto
- prevent sstate, and local_mirrors to work correctly
- does not work with corporate proxies, as http_proxy variable is only available in do_fetch

I first created #83 as a workaround for the proxy issue, but know I think it was not a good idea for the above reason.

Creating this as an issue so that it is easier for downstream user to understand the situation and help.

### LINKED ISSUE #122: how to handle krogoth branch
So I'm using krogoth and I actually need Rust 1.12.1 for my project. Unfortunately a lot of work has gone into master that makes it incompatible to just cherry-pick that version into krogoth (in fact its not possible because until recently multiple versions couldn't be available at all). There's also issue #87 which is fixed in master.

So the question is do we want to replace the krogoth branch with one based on master (effectively a merge and then reverting out the changes we've done for explicitly for morty) or should we make a `krogoth-1.12.1` branch that will address #87 and my issues. I would only support the `krogoth-1.12.1` branch and consider the older one as deprecated.

/cc @meta-rust/maintainers feedback?

cardoe:
Merges the changes in the morty branch into krogoth to get Rust 1.10.0
and Rust 1.12.1 into krogoth. Additionally this allows for Rust, Cargo
and any recipes to be built without internet access if they are defined
using cargo-bitbake. Fixes #87. Fixes #122.


cardoe:
This is functionality equivalent to the krogoth-1.12.1 branch I proposed (which is in my meta-rust fork) that I've been using. I'm switching over to using this merge branch now.

PR #132, opened 2016-12-24
rust: drop unnecessary patch

cardoe:
Not sure what version this patch appears to have become unnecessary but
I verified it by checking the output in
build/tmp/sstate-control/manifest-x86_64-rust-native.populate_sysroot
and the installed files and their paths didn't change with and without
this patch.

Signed-off-by: Doug Goldstein <cardoe@cardoe.com>


cardoe:
So I'm going through the changes aiming to port them over to Rust 1.14.0 to bring that release into the tree. I'm looking at all of our patches trying to figure out what's required if it can be upstreamed into 1.16.0 to make it easier to package that version when its released. @jmesmon can you confirm if my approach at testing if this patch is needed is valid and if it is can we drop it?

jmesmon:
Was likely only needed when we were using a non-blessed libdir for the `-cross` compiler. Since that is gone, you're right: we can drop this.

cardoe:
rebased to make it current

cardoe:
Given the two 👍 and the fact that most likely everyone's out on vacation this week I'm going to self-merge this.

PR #140, opened 2017-01-03
rust: fix cpu selection so older processors can run executables for rust-native

### LINKED ISSUE #139: Native build assumes corei7 processor extensions
## Version(s) of meta-rust
Morty (others affected as well)

Morty and morty

## Expected result
rust-native should build correctly

## Actual result
rust-native build fails with Illegal instruction

## Steps to reproduce
Attempt to build rust-native on a machine that does not have 'modern' features of a core i7.

The optimizations for rust will not execute on older intel processors because we specify a tune feature set of corei7.  This is not always the case.

CPU Info
```
processor	: 3
vendor_id	: GenuineIntel
cpu family	: 6
model		: 15
model name	: Intel(R) Xeon(R) CPU            5160  @ 3.00GHz
stepping	: 6
microcode	: 0xd2
cpu MHz		: 3000.200
cache size	: 4096 KB
physical id	: 3
siblings	: 2
core id		: 1
cpu cores	: 2
apicid		: 7
initial apicid	: 7
fpu		: yes
fpu_exception	: yes
cpuid level	: 10
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx lm constant_tsc arch_perfmon pebs bts rep_good nopl aperfmperf pni dtes64 monitor ds_cpl vmx est tm2 ssse3 cx16 xtpr pdcm dca lahf_lm tpr_shadow dtherm
bugs		:
bogomips	: 6000.09
clflush size	: 64
cache_alignment	: 64
address sizes	: 36 bits physical, 48 bits virtual
```

derekstraka:
Resolves #139

### DIFF
--- a/recipes-devtools/rust/rust.inc
+++ b/recipes-devtools/rust/rust.inc
@@ -177,7 +177,7 @@ def arch_to_rust_target_arch(arch):
 
 # generates our target CPU value
 def llvm_cpu(d):
-    cpu = d.getVar('TUNE_PKGARCH', True)
+    cpu = d.getVar('PACKAGE_ARCH', True)
     target = d.getVar('TRANSLATED_TARGET_ARCH', True)
 
     trans = {}


jmesmon:
So previously we would pass whatever `llvm_cpu` returns. Is there some reason that function returning the correct value?

It looks like `PACKAGE_ARCH` is set to `TUNE_ARCH` in meta/conf/bitbake.conf, and then is overridden in native.bbclass to be `BUILD_ARCH`. Perhaps replacing `TUNE_ARCH` (which is not overridden by native.bbclass) with `PACKAGE_ARCH` would fix things without needing to add the `native` override?

derekstraka:
I didn't look into the root cause of what change caused this to no longer function.  I had assumed an LLVM update added new optimizations, but I have no evidence of that.  Here are the set of envs that were suggested.

`bitbake -e rust-native > host.txt`
`bitbake -e rust > target.txt`

```
grep TUNE_PKGARCH= *.txt
host.txt:TUNE_PKGARCH="corei7-64"
target.txt:TUNE_PKGARCH="corei7-64"
```

```
grep TUNE_ARCH= *.txt
host.txt:TUNE_ARCH="x86_64"
target.txt:TUNE_ARCH="x86_64"
```
```
grep TRANSLATED_TARGET_ARCH= *.txt
host.txt:TRANSLATED_TARGET_ARCH="x86-64"
target.txt:TRANSLATED_TARGET_ARCH="x86-64"
```

```
grep PACKAGE_ARCH= *.txt
host.txt:PACKAGE_ARCH="x86_64"
target.txt:PACKAGE_ARCH="corei7-64"
```

Looks like `PACKAGE_ARCH` has potential.  I can try that if it is preferred.

jmesmon:
Ya, it's probably best to use `PACKAGE_ARCH` as it essentially already has the logic we'd be replicating here. It also means that `llvm_cpu` doesn't need a note about "only works for target". Lets us keep all potential remapping to llvm cpu names in 1 place.

derekstraka:
We had a brown out last night that caused three of the slaves to die mid build.  I just kicked off the build again

PR #141, opened 2017-01-27
Make builds reproducible for sstate

srwalter:
Without these commits I get various symbol / crate_hash mismatch errors when building.

It would be nice if we had a unit test to make sure we don't break sstate every time we upgrade rust.  In particular, if we built libstd-rs in two differently-named work dirs, we would have seen that the symbols have different names in each work directory, even though the bitbake task hash is the same.

### DIFF
--- a/classes/cargo.bbclass
+++ b/classes/cargo.bbclass
@@ -13,6 +13,7 @@ BASEDEPENDS_append = " cargo-native"
 
 # Ensure we get the right rust variant
 DEPENDS_append_class-target = " virtual/${TARGET_PREFIX}rust ${RUSTLIB_DEP}"
+RDEPENDS_${PN}_append_class-target = " ${RUSTLIB_DEP}"
 DEPENDS_append_class-native = " rust-native"
 
 # Cargo only supports in-tree builds at the moment
--- a/classes/force-crate-hash.bbclass
+++ b/classes/force-crate-hash.bbclass
@@ -1,0 +1,1 @@
+RUSTFLAGS += "-C crate_hash=${BB_TASKHASH}"
--- a/classes/rust-bin.bbclass
+++ b/classes/rust-bin.bbclass
@@ -1,9 +1,9 @@
 inherit rust
+inherit force-crate-hash
 
 DEPENDS_append = " patchelf-native"
 RDEPENDS_${PN} += "${RUSTLIB_DEP}"
 
-RUSTFLAGS += "-C crate_hash=${BB_TASKHASH}"
 RUSTC_ARCHFLAGS += "-C opt-level=3 -g -L ${STAGING_DIR_HOST}/${rustlibdir}"
 EXTRA_OEMAKE += 'RUSTC_ARCHFLAGS="${RUSTC_ARCHFLAGS}"'
 
--- a/recipes-devtools/cargo/cargo.inc
+++ b/recipes-devtools/cargo/cargo.inc
@@ -10,6 +10,7 @@ DEPENDS = "openssl zlib libgit2 curl ca-certificates libssh2"
 
 SRC_URI = "\
 	http://static-rust-lang-org.s3.amazonaws.com/cargo-dist/${CARGO_SNAPSHOT} \
+	file://0001-Don-t-hash-source_id.patch \
 "
 
 B = "${S}"
--- a/recipes-devtools/cargo/cargo/0001-Don-t-hash-source_id.patch
+++ b/recipes-devtools/cargo/cargo/0001-Don-t-hash-source_id.patch
@@ -1,0 +1,27 @@
+From a65784be8fdbc7dffa6d568bcc364cb67655ebfe Mon Sep 17 00:00:00 2001
+From: Steven Walter <stevenrwalter@gmail.com>
+Date: Fri, 27 Jan 2017 11:24:57 -0500
+Subject: [PATCH] Don't hash source_id
+
+This makes hashes reproducible when building directly in the source as
+well as from a registry. Reproducible file names are necessary for
+linking against externally built crates.
+---
+ src/cargo/core/package_id.rs | 1 -
+ 1 file changed, 1 deletion(-)
+
+diff --git a/src/cargo/core/package_id.rs b/src/cargo/core/package_id.rs
+index b29b8ca..7691d38 100644
+--- a/src/cargo/core/package_id.rs
++++ b/src/cargo/core/package_id.rs
+@@ -66,7 +66,6 @@ impl Hash for PackageId {
+     fn hash<S: hash::Hasher>(&self, state: &mut S) {
+         self.inner.name.hash(state);
+         self.inner.version.hash(state);
+-        self.inner.source_id.hash(state);
+     }
+ }
+ 
+-- 
+2.7.4
+
--- a/recipes-devtools/rust/libstd-rs_1.12.1.bb
+++ b/recipes-devtools/rust/libstd-rs_1.12.1.bb
@@ -32,4 +32,5 @@ do_compile_prepend () {
 do_install () {
     mkdir -p ${D}${rustlibdir}
     cp ${B}/${TARGET_SYS}/release/deps/* ${D}${rustlibdir}
+    rm ${D}${rustlibdir}/libstd*.rlib
 }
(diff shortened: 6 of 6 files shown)


derekstraka:
I believe crate hash was removed intentionally (1f188693 related to #126)

srwalter:
Thanks for pointing that out.  Indeed, cargo does not generally benefit from the crate_hash flag, as usually it just generates an executable that's not consumed by later recipes.  libstd-rs is a special case in this regard; I'll re-arrange the patch accordingly.

jmesmon:
@srwalter are there steps to reproduce the errors you're seeing?

On removing the rlib: don't most users need the rlib to link against libstd statically?

On setting the crate_hash: as @derekstraka noted, we were having issues where builds would fail if 2 crates with the same name and different versions were used. How does passing `-C crate_hash=${BB_TASKHASH}` affect cargo based builds with multiple crates? Does it override only the top-level crate? Or all of them? If all crates have the same crate_hash, why wouldn't that be a problem?

On not hashing source_id: seems like this could break builds where crate A with version B is retrieved both from a git url and from crates.io (for this example, lets say top-level crate C uses the git version and C also depends on D which uses the crates.io version). Is there some way to restrict source_id instead?

srwalter:
@jmesmon  @derekstraka pointed out, it's probably not appropriate to always pass -C crate_hash with cargo.  I reworked my patch to only do so for libstd-rs.

To reproduce the sstate problem, these steps should work:
1) Use bitbake to generate an sstate entry for an any app
2) Create a new bitbake workdir with a different path.  Ensure that the sstate entry for the app can be found, but not the sstate entry for libstd-rs
3) Use bitbake to generate an image containing the app

The app will fail to run inside the generated image.  Because the second build couldn't find an sstate entry for libstd-rs, it will have rebuilt it.  Since the workdir path is different, cargo will have used a different -C metadata argument, which means the generated libstd.so will be incompatible with the previous build.  Since bitbake *did* find an sstate entry for the app, it won't have relinked the app, so the app will still be referring to the "old" symbol names.

This is all correct and expected behavior on the part of bitbake.  The problem is that libstd-rs generates incompatible output for subsequent rebuilds.

srwalter:
@jmesmon removing the rlib doesn't harm existing cargo clients.  They will dynamically link against libstd.so if that's all that's available.  I did notice, however, that cargo clients would lack an RDEPENDS on libstd-rs, so I fixed that up as well.

cardoe:
So let me see if I understand, after this change any binary built with cargo will be dynamically linked against libstd.so? That's a regression to me.

cardoe:
Its also worth noting that I have no idea how to add your crate hash bits in 1.14.0 and as a result I've dropped them in that branch that I haven't sent as a PR. They've completely redone that area of code as they are moving towards incremental compiles.

srwalter:
@cardoe your understanding is correct.  How can you view dynamic linking as a regression?  Do you statically link every executable against libc?

cardoe:
@srwalter Well considering Rust naturally doesn't dynamically link libstd.so and that there is a flag to dynamically link then you've broken the natural behavior of the ecosystem. If you want this for certain builds feel free to add `-C prefer-dynamic`

srwalter:
@cardoe going to have to agree to disagree, there.  rust's behavior with regards to libraries and versioning and integrating with larger build systems is crappy, at best.  That's why improving it is on their roadmap for improvement in 2017.  The point of meta-rust is not to make bitbake as crappy as rust is, rather we should strive to make rust integrate nicely into bitbake.  That means not breaking sstate, that means generating and using shared objects, and that means having recipes for common intermediate libraries (not just using cargo to build the end executables).

cardoe:
Ok well then I'm a thumbs down on this merge request. Others can feel free to vote on it.

derekstraka:
Removing existing functionality in the name of preserving SSTATE functionality isn't all that helpful especially to those that are relying on these features.  There should be a thoughtful migration path forward if this does in fact break existing functionality

tylerwhall:
I think we need to define what our use cases are for meta-rust. I've only known people to use OE for building a self-contained rootfs and optionally package feeds that go along with that distro config. Upstream OE doesn't seem tailored for the case of buliding standalone portable apps, since it has no built-in mechanism for publishing individual app binaries and there doesn't seem to be any upstream consideration that the binaries built run in any environment other than the current distro configuration.

I wouldn't expect most OE users to care about static vs. dynamic as long as the executable runs in the final image. If both ways produce a working executable on the target, I would expect most to prefer dynamic for the memory/code space savings.

Formally I would define a successful Rust executable as one that runs correctly when installed as part of the target image. What is driving the requirement that the binary must be completely statically linked? That feels more like an implementation detail that's not part of the external deliverables of an OE build.

@jmesmon I'm looking into limiting the hash inputs for the source_id to be stable across workspaces, but a general solution doesn't look easy given the current structure of the code. I might be able to special-case the top-level crate's path to not include the workdir with hopefully little tear-up, though.

derekstraka:
That may be an implementation decision but removing the *.rlib files altogether takes away any freedom to decide the implementation.

derekstraka:
Another item to point out is that rust refactored the code that implements the crate_hash (https://github.com/rust-lang/rust/commit/06242ff15d2b510689a99c33aa2495da11a75fad).  @cardoe is on the verge of updating to 1.14 which would render that unusable without a fair bit of effort.  Given that it causes #126 the prevailing thought was to drop the patch in the interim.

derekstraka:
If we are going down the path of removing static linkage from the mainline, I think the rust packages also need to be refactored to put the *.rlib files in the static-dev or -dev packages rather than removing them entirely.

srwalter:
@derekstraka whether or not this removes functionality is what I'm trying to establish.  With these changes, I can build a rootfs image and run rust code in that image.  I don't understand the use case @cardoe has in mind where a dynamic library for libstd is apparently unacceptable.  I do know that without these changes, it only works in the static case.  If I build against libstd with -C prefer-dynamic, I get a compiler error along the lines of "multiple crates found: $SYSROOT/libstd.so $SYSROOT/libstd.rlib".  So it would seem we must choose between static linking and dynamic linking, at least for libstd.  For the use case of building a root filesystem for an embedded device with multiple rust apps where code size matters, dynamic linking is clearly preferable.  But without understanding the use case that this breaks, we can't offer an alternative or weigh the costs and the benefits.

@cardoe, do you care to elaborate on the use case you have in mind that would be harmed by this?

As for rust 1.14, we'll cross that bridge when we come to it.  If you have to drop the crate_hash patch to do the upgrade, so be it.  Upstream has been making strides to make the crate_hash more deterministic, so there's at least some chance that in 1.14 we won't need the patch, anyway.

cardoe:
So fundamentally my issues here are:

1. The cargo patch has not been submitted upstream and doesn't have an associated issue upstream which is really what the goal of this repo is because we need the ability to keep up with releases and the more downstream patches we have the harder it is. Fundamentally the patch is to keep your SSTATE good but makes bad assumptions. Firstly Rust does not have a stable ABI between generated .so files so one .so build isn't the same as another necessarily. This change attempts to make that happen. One way they try to catch breakage is by including the path and everything with it in there, but this removes that. For us we use a mix of containers, vagrant, and puppet to ensure all of our build environments are identical so that one .so is really compatible with another. But it sounds like you are not using something like that in your environment so this attempts to paper over it. It could cause issues down the road for other people using SSTATE or it could bite people similar to #126.
2. This universally takes control away from people attempting to use this layer by forcing everyone to link dynamically to libstd when there is an existing way to do so that's not used.
3. Many assumptions that Yocto is only for small embedded systems when it is much more now days and is really a meta distribution and has many use cases and as such so does Rust.

posborne:
I will chime in and say that I currently have a use case for statically linked rust binaries generated by Yocto.

In my case, we generate an initial update package (we do full image updates) that includes a fetcher written in Rust.  This fetcher is currently a statically linked Rust binary that we build for the target using a Yocto recipe.

We actually build this initial update image as its own image, so it would be possible, but far from preferable, to include other dynamic libraries in with the generated executable and make sure that the linker loader can find those libraries at runtime.

Currently, we are using `meta-rust-bin` which is far less sophisticated than `meta-rust` (it does not support dynamic linking at all, for instance) but also pretty predictable.

srwalter:
I'll rebase this once rust 1.14 and cargo 1.15 have landed.  That'll also give some time for @tylerwhall 's upstream cargo fix to get some attention:  https://github.com/rust-lang/cargo/pull/3611

PR #142, opened 2017-01-29
bump to Rust 1.14.0

cardoe:
This brings in Rust 1.14.0 which should become the legacy version of pyro and the stable version in morty. See https://github.com/meta-rust/meta-rust/wiki#supported-versions for more details.

Cargo 0.15.0 will come as a separate PR.

These files are all copied from their 1.12.1 versions with the following exceptions:

- rust-source-1.14.0.inc - this had RS_KEY dropped and hashes/versions modified
- libstd-rs_1.14.0.bb - RUSTC_BOOTSTRAP_KEY renamed to RUSTC_BOOTSTRAP
- The 3 Rust source patch files were rebased against Rust 1.14.0 code base
- The addition of `-C crate_hash` was dropped due to the fact that Rust 1.14.0 has started to add incremental compilation and has entirely changed the behavior. That change needs to be rethought out in the new architecture and reimplemented.


srwalter:
Changes look good to me.  I'll investigate whether -C crate_hash is still needed, and figure out how to handle it if so.

cardoe:
@tylerwhall That's a good point. Let me refactor that bit and try to figure out why gcc is ICEing on the machines @derekstraka and I have running as the test box. Its working locally but I don't like that its not working in the test environment.

cardoe:
I've re-implemented this as suggested by @tylerwhall and I think this will improve version bumps in the future. My want is to make version bumps painless to the point that I can bump Gentoo and meta-rust on the day of the release.

cardoe:
The reason why Cargo 0.15.0 isn't in here is that the git fetcher from Yocto is complaining about the commit that 0.15.0 is tagged because I believe it was done from a branch that was integrated back into master after the 0.15.0 release.

jmesmon:
@cardoe on cargo 0.15.0: setting `nobranch=1` or `branch=origin/rust-1.14.0` might work.

lgtm anyhow.

PR #145, opened 2017-01-31
rust-llvm.inc: work around build race

srwalter:
The way LLVM builds its native utilities is not safe (it runs a
recursive cmake in the same work directory as the main instance).  To
prevent those instances from stepping on each other, we'll build the two
native utilities separately before starting the main build.

Fixes issue #143

### DIFF
--- a/recipes-devtools/rust/rust-llvm.inc
+++ b/recipes-devtools/rust/rust-llvm.inc
@@ -80,6 +80,8 @@ do_compile_prepend_class-target() {
 }
 
 do_compile() {
+    oe_runmake NATIVE_LIB_LLVMTABLEGEN
+    oe_runmake NativeLLVMConfig
     oe_runmake
 }
 


derekstraka:
@srwalter - Any idea if there is an upstream issue for this?  I've never seen this issue before and we run with a lot more jobs (-j) than the usual user probably would (between 48-64).  Must just be a sweet spot for hitting this.

derekstraka:
This would need backport to `morty` and probably `krogoth` once it goes into master

srwalter:
@derekstraka I wasn't able to find an upstream issue for this.  It will only occur when building support for non-native architectures and not using an already-installed LLVM.  Another possible fix would be to have another recipe that builds these native LLVM tools, then point rust-llvm-native at those.

The build machines where I've seen this fail are 32 core machines, so running with -j64.  I've only seen it fail a couple of times out of dozens of builds, however.

I'll post merge requests for backports once everyone is happy with this one

cardoe:
@srwalter if you want direct commit this to morty and krogoth as well I think that's fine.

PR #149, opened 2017-02-14
Add the ability to use and store build cache artifacts

derekstraka:
Before this PR, all builds would run without and use of SSTATE cache.  With this addition, the build times can be significantly reduced.


derekstraka:
Took the ~90 min build down to ~3 min when using SSTATE cache:

https://ci.asterius.io/job/meta-rust/job/PR-149/

PR #150, opened 2017-02-14
Add the ability to use and store build cache artifacts

derekstraka:


scripts/publish-build-cache.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+if [[ $# -lt 1 ]]; then
+    echo "No Yocto branch specified, defaulting to master"
+    echo "To change this pass a Yocto branch name as an argument to this script"
+fi
+branch=${1-master}
+
+rsync -avz -e "ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" --progress build/downloads yocto-cache@build-cache.asterius.io:/srv/yocto-cache/
```
srwalter: How beneficial is it to cache the downloads directory?

derekstraka: Not as much for this project, but I'm also building base images for meta-virtualization and meta-measured.  Our cache is in AWS and it ends up being a significant performance boost on some of the mirrors (kernels especially).

Issue #113, opened 2016-11-17
libstd-rs fails with "libcompiler-rt.a: ...... recompile with -fPIC" when building for MIPS


Fulkerson:
I'm trying to make meta-rust work with a mips32r2 little endian target. Things are progressing nicely and I've managed to build rust properly.

However the libstd-rs recipe failes with the following error:

    ....../usr/lib/libcompiler-rt.a(mulodi4.o): relocation R_MIPS_26 against `__divdi3@@GLIBC_2.0' can not be used when making a shared object; recompile with -fPIC

I did as told and added `-fPIC` to `COMMON_INCLUDES=` in _compiler-rt/make/config.mk_ and now things work well enough. I am able to build and run the rust-hello-world recipe on target.

Does anyone have an idea why -fPIC is not added by default. Should it? If it should, what is the best practice? I don't think COMMON_INCLUDES is the proper place after all.

Fulkerson:
Did some more digging. It seems as though -fPIC was removed from compiler-rt in commit a4b1f78283db06dc5bbec6abaab7a93086697c44.

"Rust: Silence -fPIC warnings on Windows"

However it is back again in at least version 1.12.1 of rust, so I think this will solve itself in future versions of rust.

cardoe:
@Fulkerson Are you using master or one of the other branches? Do you have any proposed patches?

Fulkerson:
I am using master. My solution was to simply revert commit a4b1f78283db06dc5bbec6abaab7a93086697c44 of rust. However it is already fixed upstream so when rust is bumped it should fix itself.

meta-rust does not currently work for MIPS anyway. I plan on making a pull request when stuff seem to work properly!

jmesmon:
morty & master (pyro) now ship 1.12.1

derekstraka:
@Fulkerson - Do you have a branch where the changes for MIPS live?  I can touch things up for 1.14 if you have a starting point and add it to the CI build.

PR #151, opened 2017-02-25
Add support for mips32r2el-nf

Fulkerson:
This patch was requested in issue #113.

With this I am able to build a rust compiler for a mips32r2 LE target without any floating point unit. It seems to compile proper binaries, although I have not tested it extensively.

This is mostly a quick hack and there are a few dirtinesses. The mips targets are a bit annoying as the default targets in rustc are not very consistent. For example:

mipsel_unknown_linux_gnu -> Little endian mips32 hard float (though no FPU selected)
mipsel_unknown_linux_gnu -> Little endian mips32 soft float
mips_unknown_linux_gnu -> Big endian mips32**r2** hard float (though no FPU selected)
mips_unknown_linux_gnu -> Big endian mips32**r2** soft float

This is only an issue in the mk/ files. The target file is fairly straight forward except "generic" is not a target llc lists for the mips architecture. I don't know exactly how compatible the different mips cpus are so I think it would be best to keep them separated.


derekstraka:
Are you using qemu to validate?  If so, we should add the target to the Jenkinsfile so the CI builder has MIPS too.

recipes-devtools/rust/rust.inc
```diff
@@ -192,8 +217,27 @@ def llvm_cpu(d):
...
+    if "mips32r2" in feat:
+        return "mips32r2"
+    elif "mips32" in feat:
+        return "mips32"
+
+    # TODO: Proper x86 targets. "generic" is a valid choice.
+    # TODO: Proper ARM cpu selection, for now "generic" is a valid cpu choice.
+    return "generic"
+
+TARGET_LLVM_CPU="${@llvm_cross_cpu(d)}"
```
tylerwhall: Looks like this makes all x86 targets "generic" where there were previously some specific options. Is that what we want?

jmesmon: Unlikely. Keeping & extending `llvm_cpu()` and setting `TUNE_FEATURES_class-native = ""` or something is likely a better choice. `TUNE_FEATURES` isn't cleared by class-native, so we only want to examine it in the cross & target cases. `PACKAGE_ARCH` and `TRANSLATED_TARGET_ARCH`, however, are modified in class-native, so we can keep using them in all cases (as `llvm_cpu()` currently does).

recipes-devtools/rust/rust.inc
```diff
@@ -315,6 +366,10 @@ def rust_gen_mk_cfg(d, thing):
         '-e', '/^CFG_GCCISH_CXXFLAGS/ s;$; {};'.format(cxxflags_for(d, thing)),
         '-e', '/^CFG_GCCISH_LINK_FLAGS/ s;$; {};'.format(ldflags),
 
+        # The mipsel target specify flags for the mips32. Replace them with
+        # mips32r2 if that's what we use
+        '-e', 's/mips32/mips32r2/g' if is_mips32r2 else '',
```
jmesmon: I'm not entirely comfortable with this type of substitution. All the existing ones explicitly seek out variables to set, this should too.

On ldflags: are the ones that yocto provides in `LDFLAGS`, `HOST_CC_ARCH` and `TOOLCHAIN_OPTIONS` not already sufficient? Or is the goal here to remove some bad ones?

Fulkerson: I agree it is not very nice.

The flags from yocto are fine, they specifiy the cpu (e.g. -march=34kc). The problem is the mk/cfg/ files specifies -mips32 (equivalent to -march=mips32) for CFG_JEMALLOC_CFLAGS, CFG_GCCISH_CFLAGS, CFG_GCCISH_LINK_FLAGS. But the -march=34kc flag is placed after the -mips32, so perhaps that will always be chosen. In that case this substitution is unnecessary. Do you know if that is the case?

Here's an example of what I mean:
`CFG_GCCISH_CFLAGS_mipsel-poky-linux := -Wall -g -fPIC -mips32 -mabi=32  -mel -mabi=32 -msoft-float -march=34kc -msynci --sysroot=...`

However these files are completely gone by now so perhaps it would be better to put effort into bumping to rust 1.15

Fulkerson:
@derekstraka I have not used the qemumips target. I think it is a big endian target which wouldn't work with these changes, but I will try to add big endian support as well. I think it should be trivial to add a qemumips le machine for testing in the meta-rust layer as well if that would be acceptable.

recipes-devtools/rust/rust.inc
Fulkerson: I'm not quite sure I understand. PACKAGE_ARCH is always set to x86_64 (note the underscore) for me, no matter if I'm building rust-cross-mipsel or rust-native. I assume this is because rust-cross-mipsel is still a x86_64 package. I'm not sure that PACKAGE_ARCH will ever result in anything usable by the trans["..."] dict.

TRANSLATED_TARGET_ARCH is pretty boring as it will result in something like "mipsel" for my target which isn't really enough to do anything.

recipes-devtools/rust/rust.inc
cardoe: I've got a tracker for rustbuild at #152

recipes-devtools/rust/rust.inc
jmesmon: Hmm, I'll take a look at the `PACKAGE_ARCH` bits in detail again when I have a chance, I had thought it did what we wanted here. That said, it does look like it's getting set to BUILD_ARCH by cross.bbclass, which is almost certainly not what we want.

Note that the rust-cross targets don't actually build anything at this point, they're only cross in the sense that they drop files in the native sysroot but need some details about the target to generate them.

Solutions here are a bit tricky due to the way json files are generated. Likely could be made better by only generating 1 per recipe (instead of trying to do all of HOST/BUILD/TARGET).

recipes-devtools/rust/rust.inc
jmesmon: In gcc (and many other command line interfaces with options), later arguments override earlier ones, with the intent of supporting cases like this one. Appending the flag should work fine unless `-mips32` somehow implies something slightly different than a `-march` option.

Another option here is just to have meta-rust generate the full set of cflags, which I have considered in the past. Trying to build on top of the `mk/cfg` stuff instead of generating it completely is a bit funky (but it was quick to get working).

cardoe:
@Fulkerson Would you be interested in rebasing this now?

Fulkerson:
Dropped in favor of #182

Issue #152, opened 2017-03-07
switch to using rustbuild


cardoe:
rustbuild is the new Cargo driven system to build Rust. Upstream has declared this build system to be complete with the 1.15 release. Preferred with the 1.16 release and with 1.17 the makefile based build system has been removed leaving only rustbuild.

So we need to start taking the steps to move towards rustbuild

PR #154, opened 2017-03-12
[WIP] Initial attempt at using rustbuild.

Fulkerson:
This is just a work in progress trial and should not be considered for merging!

I've tried to adapt the rust recipe for using [rustbuild](https://github.com/rust-lang/rust/blob/master/src/bootstrap/README.md) (see #152).

I use the `config.toml` approach instead of `./configure` because the configure script has dependencies to the `mk/` directory which does not exist in rust master and works very badly for our special targets. Also it is very straightforward.

So far I've gotten to the point where rust-native starts running cargo to build but it fails when trying to fetch the gcc crate. I suppose rust requires the following crates in order to build:
https://github.com/rust-lang/rust/blob/master/src/bootstrap/Cargo.toml

Any ideas how to fetch the rustbuild cargo dependencies from within openembedded? Could we perhaps use _cargo.bbclass_ and _crate-fetch.bbclass_ somehow?

Finally I believe we need to migrate our flags from  `rust_gen_mk_cfg` to the [gcc-rs way](https://github.com/alexcrichton/gcc-rs#external-configuration-via-environment-variables)

### DIFF
--- a/recipes-devtools/rust/rust.inc
+++ b/recipes-devtools/rust/rust.inc
@@ -10,6 +10,12 @@ SRC_URI = "\
 	https://static.rust-lang.org/dist/${RUST_SNAPSHOT}.tar.gz;name=rust-snapshot \
     "
 
+# rustbuild uses cargo to build rustc.
+require recipes-devtools/cargo/cargo-snapshot.inc
+SRC_URI += "\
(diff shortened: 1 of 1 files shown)


jmesmon:
On flags: cargo.bbclass has a `oe_cargo_fix_env` that likely does what is needed, it could be moved to some shared location.

Fulkerson:
Abandoned in favor of #158.

Issue #138, opened 2017-01-03
use RUST_TARGET_PATH instead of patching rustc


cardoe:
Instead of patching rustc to load the proper target for us. We should use the environment variable to ensure that the compiler gets the right target. This will lead to more consistent data when using shared SSTATE between different machines.

see japaric/xargo#44 and rust-lang/rust#24666

jmesmon:
The existing patch to rustc for locating targets was really designed just to add a default value to RUST_TARGET_PATH. Should be possible to simply drop the patch & always set RUST_TARGET_PATH to the path that would have been patched in.

But based on the mention of the sstate stuff, I'm guessing you want to use multiple RUST_TARGET_PATH elements to allow the yocto-target libstd & target.json to move to a location outside of the yocto-host sysroot (which still having the host libstd & target.json in the host sysroot)? Or something along those lines?

cardoe:
I was more thinking of your first paragraph. I hadn't really thought about moving things out of the host sysroot yet.

cardoe:
I was trying to use rustbuild and ran into something where the path was getting baked in when `--target` was a full path. I know that the `crate_hash` bits exist to try and neutralize some of the sstate changes due to paths so I thought this might help all around.

jmesmon:
We're not currently passing `--target` a full path, only `${HOST_SYS}` (as far as I can see). Even for the bootstrapping (where we're already using `RUST_TARGET_PATH`).

Is there something fiddly about `rustbuild` that is causing a full path to be passed?

cardoe:
This is now finished.

PR #161, opened 2017-06-21
drop custom compiler patch to provide target

cardoe:
This drops the custom compiler patch to allow us to find the target thereby fixing #138. It also changes how Cargo is built to not use the configure script with make since that is not compatible. It has some improvements to the debug output and lastly provides C++ wrappers to the build process. Lastly it converts the Cargo recipe to one generated by `cargo-bitbake`.


cardoe:
@Fulkerson I committed some of these patches with you as the author since I got them from your series.

recipes-devtools/cargo/cargo_0.16.0.bb
```diff
@@ -1,85 +1,102 @@
...
+crate://crates.io/winapi-build/0.1.1 \
+crate://crates.io/winapi/0.2.8 \
+crate://crates.io/ws2_32-sys/0.2.1 \
 "
 
-SRCREV_cargo = "6e0c18cccc8b0c06fba8a8d76486f81a792fb420"
 
-S = "${WORKDIR}/git"
 
+# FIXME: update generateme with the real MD5 of the license file
```
derekstraka: Can probably drop this line.  Are the other licenses no longer valid?

cardoe: So `cargo_0.16.0.bb` is purely auto-generated. So yes I can drop that line but its a manual edit until I pull it out of `cargo-bitbake`. The other licenses are valid but not able to be picked up by `cargo-bitbake` for two different reasons. They're listed in `cargo.inc` currently.

cardoe: I'll squash the drop of this line in and I'll also toss on a commit that converts `rustfmt` over to being generated by `cargo-bitbake`.

PR #165, opened 2017-09-28
Switch to using rustbuild

### LINKED ISSUE #152: switch to using rustbuild

cardoe:
Switch to using rustbuild using the work of @Fulkerson in #158 and should close #152.


cardoe:
I tried to address most of my original review comments after receiving some feedback on others. I found some other issues that I attempted to address as well. While the changes likely aren't perfect, I think they're good enough. Rather than let perfect be the enemy of good I'm trying to bring in the newer build system to bring in the newer version.

cardoe:
Rebased so it should pass on master. I forgot that locally I hadn't upgraded my meta to new enough to be affected by the change since that would have required a bigger rebuild locally. I was just looking to develop quicker and not stare at a compile.

PR #166, opened 2017-09-29
Fix the bblayers sample file based on poky renames

derekstraka:


derekstraka:
This should fix the build issues you see with #165

Issue #162, opened 2017-06-28
libstd-rs build fails on morty


SanchayanMaity:
## Version(s) of meta-rust
morty Commit Tag: f5387107af7fdb4558d1886d6f68da5ab8403a48

## Version(s) of poky and/or oe-core
oe-core morty

## Expected result
Build of a simple rust hello world project to succeed

## Actual result
Build failure of libstd-rs in the process of building hello world project

## Steps to reproduce
1. Used the below bitbake recipe generated with cargo bitbake to build a simple hello world program for a iMX6 ARM target

```
# Auto-Generated by cargo-bitbake 0.3.6
#
inherit cargo

# If this is git based prefer versioned ones if they exist
# DEFAULT_PREFERENCE = "-1"

# how to get hello_cargo could be as easy as but default to a git checkout:
# SRC_URI += "crate://crates.io/hello_cargo/0.1.0"
SRC_URI += "git://github.com/SanchayanMaity/hello_world_metarust.git;protocol=https"
SRCREV = "04660d0bfce25c2c4c45ec35b47edc56ac0249be"
S = "${WORKDIR}/git"
CARGO_SRC_DIR=""
PV_append = ".AUTOINC+04660d0bfc"

# please note if you have entries that do not begin with crate://
# you must change them to how that package can be fetched
SRC_URI += " \
"



# FIXME: update generateme with the real MD5 of the license file
LIC_FILES_CHKSUM=" \
file://GPL-3.0;md5=generateme \
"

SUMMARY = "Hello world OE meta-rust test"
HOMEPAGE = "sanchayanmaity.github.io"
LICENSE = "CLOSED"

# includes this file if it exists but does not fail
# this is useful for anything you may want to override from
# what cargo-bitbake generates.
include hello_cargo.inc

```
2. Build fails with the following error log

```
ERROR: libstd-rs-1.15.1-r0 do_compile: Function failed: do_compile (log file is located at /build/mas/morty/build/tmp-glibc/work/armv7at2hf-vfp-angstrom-linux-gnueabi/libstd-rs/1.15.1-r0/temp/log.do_compile.31966)
ERROR: Logfile of failure stored in: /build/mas/morty/build/tmp-glibc/work/armv7at2hf-vfp-angstrom-linux-gnueabi/libstd-rs/1.15.1-r0/temp/log.do_compile.31966
Log data follows:
| DEBUG: Executing shell function do_compile
[...]

PR #167, opened 2017-09-29
layer.conf: add the openembedded-layer to the depends since we use libgit2 from there

derekstraka:
Found using the yocto compat script.


cardoe:
Could this fix #162?

cardoe:
@derekstraka Does this address #162 and do we need to backport?

PR #168, opened 2017-09-29
bump to Rust 1.20.0 and Cargo 0.21

cardoe:


Fulkerson:
I've gotten this to work. I don't know if I can attach my patches to a pull request on github but there are two commits here:

https://github.com/Fulkerson/meta-rust/commits/rust-1.20.0

What has happened is the system allocator is now the default in rust  although it is overridden in linux_base which we don't use as we generate a JSON target. By leaving exe-allocation-crate out the default is used.

This is done by a rather ugly version compare in rust_gen_target but we will be able to remove that as soon as we drop support for older rust compilers so I don't think we should waste any time on that.

One thing that is added in rust for ARM targets is a ABI blacklist. Things work fine without it but we should probably add it in the future. If I understand things correctly the compiler will fail if you try to use an ABI for an architecture without the support so it doesn't seem that important.

derekstraka:
Close in favor of #169

PR #169, opened 2017-09-30
bump to Rust 1.20.0 and Cargo 0.21

Fulkerson:
This is pull request #168  by Cardoe with two small commits that make it build.

What has happened is the system allocator is now the default in rust although it is overridden in linux_base which we don't use as we generate a JSON target. By leaving exe-allocation-crate out the default is used.

This is done by a rather ugly version compare in rust_gen_target but we will be able to remove that as soon as we drop support for older rust compilers so I don't think we should waste any time on that.

Another thing that has changed in rust for ARM targets is an ABI blacklist. Things work fine without it but we should probably add it in the future. If I understand things correctly the compiler will fail if you try to use an ABI for an architecture without the support so it doesn't seem that important.


cardoe:
Thanks for the update @Fulkerson. I pushed what I had as I walked out the door on vacation. I knew I was concerned about the original allocator hardcoding in the original PR for a reason just not the right reason.

PR #170, opened 2017-10-03
cargo.bbclass: Avoid bashism in cargo_do_install

willnewton:
The double square brackets are an extension to POSIX so are not
available in some shells. Use a case statement instead.


classes/cargo.bbclass
```diff
@@ -46,15 +46,20 @@ cargo_do_compile () {
...
 			install -m755 "$tgt" "${D}${rustlibdir}"
 			have_installed=true
-		elif [ -f "$tgt" ] && [ -x "$tgt" ]; then
-			install -d "${D}${bindir}"
-			install -m755 "$tgt" "${D}${bindir}"
-			have_installed=true
-		fi
+			;;
+		*)
+			if [ -f "$tgt" ] && [ -x "$tgt" ]; then
```
cardoe: Technically you didn't change this but `-f` is redundant with `-x`

PR #177, opened 2017-10-31
support Rust 1.21.0 and Cargo 0.22.0

jcreekmore:


derekstraka:
At this point, probably just remove 1.20 in the same PR?

PR #178, opened 2017-11-03
rust-bin.bbclass fixes

srwalter:
We finally got around to upgrading to rust 1.20/1.21 internally.  We had to make some changes to rust-bin.bbclass to get everything working again.

@meta-rust/maintainers


srwalter:
Seems to me that these changes actually did build successfully, and then Jenkins crapped out for an unrelated reason after the build...

cardoe:
@srwalter Agreed.

PR #180, opened 2017-11-07
Make rustc output reproducible

srwalter:
With these changes, we're getting compatible output from the compiler with stable crate hashes even across build machines and workspaces.  This supercedes the old -C crate_hash compiler flag that we had previously used to ensure compatibility.

The patch to fix path remapping is a backport of an already-upstreamed fix, so that patch should not need to be rebased in the next upgrade.


cardoe:
Well I reviewed the patch to allow unstable options and that looks good. Great job coming up with an upstream solution and making it happen. Looks like this will close #126 for rocko and forward.

PR #181, opened 2017-11-26
rust: specify target-c-int-width

Fulkerson:
The "target_c_int_width field is added to librustc_back since rust 1.22.0 because not all rust targets (e.g. msp430-none-elf, avr) have 32-bit int types any more.

This will probably be needed when updating to rust 1.22+


srwalter:
Your commit message is a little sparse.  Could you elaborate on what problem you're fixing?

Fulkerson:
I've updated the commit message. I'm not really fixing anything other than making sure we stay up to date with how the rustc backends look. I noticed the change while doing some other work.

I don't think it will affect the current builds, but from 1.22 it is likely a required field.

srwalter:
Makes sense, thanks

PR #184, opened 2018-01-23
rust-llvm: remove custom do_compile

bkylerussell:
This works out in recent poky releases because cmake_do_compile ends up
in oe_runmake via base_do_compile, but this changed recently to allow
cmake's generic `cmake --build` feature to call the appropriate
generator instead of always assuming make.

http://git.yoctoproject.org/cgit/cgit.cgi/poky/commit?id=579d17ba5f722edb9fe79682480c4a9508ad0ed4


tylerwhall:
It looks like the build was successful but hung at the end. Also I think not having this change means we're broken against poky/master and #185 failed because of it. So I'm going to pull.

PR #187, opened 2018-02-22
Add the 1.24.0 rustc and 0.25.0 cargo to pyro.

jcreekmore:


derekstraka:
@jcreekmore - Are you going to target rocko and master branch once this lands in pyro?

jcreekmore:
I can. I didn't need anything other than pyro at the moment, but it shouldn't be too difficult to retarget this to rocko or master.

cardoe:
I would recommend addressing what the layer supports in the wiki as well. https://github.com/meta-rust/meta-rust/wiki

cardoe:
I suppose since Rust has stabilized a bit with their build system it doesn't require incompatible changes to the common bits requiring the removal of the old version. It should be possible for other users to pin at 1.20 if that's needed. There was someone in the past that opened up a bug when we moved the top end version up which is what led to the creation of the legacy/stable policy to begin with.

jcreekmore:
I can update the wiki when/if it goes in. The rebase onto rocko wasn't bad, so I am building that now.

derekstraka:
master points to pretty much the same commit IIRC

jcreekmore:
Pretty close.

derekstraka:
Two nits and then I'm good merging it in.

jcreekmore:
Which are?

derekstraka:
The two comments above.  Clearly github didn't do well.  It was tabs vs spaces and one file lacked a newline

jcreekmore:
Updated to add the newline at the end of the rust-cross_1.24.0.bb file and changed the tabs for spaces in the cargo-0.25.0.inc file.

derekstraka:
Should we close this one?

jcreekmore:
Yeah, I am going to be abandoning pyro in a week or so anyway.

PR #188, opened 2018-02-23
Add rust version 1.24.0

jcreekmore:
This should be mergeable to master as well.


derekstraka:
Errors on qemu-mips

```[qemumips] ERROR: libstd-rs-1.24.0-r0 do_compile: Function failed: do_compile (log file is located at /var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0/temp/log.do_compile.25518)
[qemumips] ERROR: Logfile of failure stored in: /var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0/temp/log.do_compile.25518
[qemumips] Log data follows:
[qemumips] | DEBUG: Executing shell function do_compile
[qemumips] | NOTE: cargo = /var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0/recipe-sysroot-native/usr/bin/cargo
[qemumips] | NOTE: rustc =
[qemumips] | NOTE: cargo build -v --target mips-poky-linux --release --features 'panic-unwind'
[qemumips] | error: failed to run `rustc` to learn about target-specific information
[qemumips] | 
[qemumips] | Caused by:
[qemumips] |   process didn't exit successfully: `rustc - --crate-name ___ --print=file-names -L /var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0/recipe-sysroot/usr/lib/rust -Zremap-path-prefix-from=/var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0 -Zremap-path-prefix-to=/usr/src/debug/libstd-rs/1.24.0-r0 -L /var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0/recipe-sysroot/usr/lib --target mips-poky-linux --crate-type bin --crate-type dylib --crate-type rlib` (exit code: 101)
[qemumips] | --- stderr
[qemumips] | error: Error loading target specification: Field target-c-int-width in target specification is required
[qemumips] |   |
[qemumips] |   = help: Use `--print target-list` for a list of built-in targets
[qemumips] | 
[qemumips] | 
[qemumips] | WARNING: /var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0/temp/run.do_compile.25518:1 exit 101 from '"cargo" build -v --target mips-poky-linux --release --features 'panic-unwind' "$@"'
[qemumips] | ERROR: Function failed: do_compile (log file is located at /var/build/build/tmp/work/mips32r2-poky-linux/libstd-rs/1.24.0-r0/temp/log.do_compile.25518)
[qemumips] NOTE: recipe libstd-rs-1.24.0-r0: task do_compile: Failed
[qemumips] ERROR: Task (/var/build/poky/meta-rust/recipes-devtools/rust/libstd-rs_1.24.0.bb:do_compile) failed with exit code '1'
[qemumips] NOTE: Tasks Summary: Attempted 670 tasks of which 66 didn't need to be rerun and 1 failed.```

**Comment by jcreekmore** [comment]
Whoops. That was a bad rebase on my part. That should be fixed in the latest version. Also, I went ahead and pre-emptively address the lack of newline at the end of the rust-cross_1.24.0.bb file and changed the tabs for spaces in the cargo-0.25.0.inc file.

**Comment by derekstraka** [comment]
@agherzan - Any concerns since you're actively using rocko

File: recipes-devtools/rust/files/rust-1.24.0/0001-librustc-always-allow-unstable-options.patch [hunk 1]
```diff
@@ -0,0 +1,43 @@
+From 6a82f31d21ac7b85211e580585cc73ab2bdb0bc9 Mon Sep 17 00:00:00 2001
+From: Tyler Hall <tyler.hall@lexmark.com>
+Date: Sun, 29 Oct 2017 16:29:03 -0400
+Subject: [PATCH] librustc: always allow unstable options
```
agherzan: Can we have Upstream status here? And SOB as well.

recipes-devtools/rust/libstd-rs.inc
```diff
@@ -20,5 +20,6 @@ do_compile_prepend () {
 do_install () {
     mkdir -p ${D}${rustlibdir}
 
+    rm ${B}/${TARGET_SYS}/release/deps/*.d
```
agherzan: We could use a comment here to explain why we do this.

recipes-devtools/rust/rust-llvm_1.24.0.bb
```diff
@@ -0,0 +1,16 @@
+require rust-source-${PV}.inc
+require rust-llvm.inc
+
+LIC_FILES_CHKSUM = "file://LICENSE.TXT;md5=e825e017edc35cfd58e26116e5251771"
+
+do_install_prepend () {
+	# the install does a sed on this without installing the file
```
agherzan: If we don't need it, can't we patch the package? It would be easier to follow up with updates.

recipes-devtools/rust/rust-snapshot-1.24.0.inc
```diff
@@ -0,0 +1,26 @@
+# Specifics for Rust 1.21.0
```
agherzan: Old version remained here.

agherzan:
@derekstraka I didn't request changes as I didn't want to block the PR but find above by review.

recipes-devtools/rust/files/rust-1.24.0/0001-librustc-always-allow-unstable-options.patch
jcreekmore: There is no upstream status here at all. As far as I know, this patch was never submitted upstream. The only reason I carried it forward from the Rust 1.21 build is that the Rocko has `-Zremap-path-prefix-from` added to the Rust flags for the patched that only works with the patched 1.21 compiler. I would be more than happy to rip that and the passing of the unstable flag out of the `rust-common.bbclass`, but I figured others would be unhappy about that.

recipes-devtools/rust/libstd-rs.inc
jcreekmore: I can add a comment in the file for this. Until I get that added,  with the incremental build support added in 1.24, the libstd deps directory also includes dependency files that get installed. Those are really only needed to incrementally rebuild the libstd library itself and don't need to be installed.

recipes-devtools/rust/rust-llvm_1.24.0.bb
jcreekmore: To me, it is going to be carrying a patch forward vs doing the remove in after the install stage. Neither really seems worse one way or the other to me; I kept the remove after the install stage to be consistent with previous versions of the recipe.

recipes-devtools/rust/files/rust-1.24.0/0001-librustc-always-allow-unstable-options.patch
agherzan: Well in that case we should use an upstream status as innapropiate as this sounds like a specific configuration we decided upon.

recipes-devtools/rust/rust-snapshot-1.24.0.inc
jcreekmore: I can fix that.

recipes-devtools/rust/rust-llvm_1.24.0.bb
agherzan: Makes sense.

recipes-devtools/rust/libstd-rs.inc
agherzan: Got it now.

recipes-devtools/rust/files/rust-1.24.0/0001-librustc-always-allow-unstable-options.patch
cardoe: You're only allowed to use "unstable" options on a nightly compiler because Rust guarantees that the compiler CLI will remain stable. The goal was to make reproducible builds which were needed by @srwalter and @tylerwhall. It was merged by them in #180. Effectively this means that you can use "unstable" options with the stable compiler generated by these recipes. This change will never be accepted upstream.

cardoe: I would say some of your review comments are issues with the #180 PR and you'd need to reach out to them to get a SOB. @jcreekmore isn't able to provide that. He's copying this patch into a named directory for Rust 1.24.0 and could instead refer to the old one instead of updating it.

agherzan: This explanation should reside in the patch. As well I propose to add an Inappropriate Upstream status as details in http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines .

jcreekmore:
@agherzan re: specifically the TARGET_C_INT_WIDTH change. When I added the patches for pyro, I followed the variable naming that was in the target file `target-c-int-width`. When I ported the commit to Rocko, the 1.21 patchset had already brought it in as `TARGET_INT_WIDTH`, so I fixed up the 1.24 changes to match that.

recipes-devtools/rust/files/rust-1.24.0/0001-librustc-always-allow-unstable-options.patch
cardoe: I don't disagree with you. None of the people involved right now had anything to do with that PR however.

agherzan:
@jcreekmore I understand now. Sounds good but this kind of explanation still needs to reside in the git log. BTW, is this PR already in master?

recipes-devtools/rust/files/rust-1.24.0/0001-librustc-always-allow-unstable-options.patch
cardoe: You could argue we could drop the patch until it meets guidelines. But I would say you'd want to update the project guidelines at https://github.com/meta-rust/meta-rust/wiki

cardoe:
I'd agree that this should go in master first and then pluck back to rocko.

recipes-devtools/rust/files/rust-1.24.0/0001-librustc-always-allow-unstable-options.patch
agherzan: No, I understand now that these were carried on from an older version. Initially I was missing that bit. So given this situation I would say leave this as it is and hopefully somebody would address it in a follow up PR. Because otherwise we will be dragging patches without any understanding and somebody reading an update PR would be completely out of context.

jcreekmore: OK, I will leave it as it, then.

jcreekmore:
I can make it go into master first. Let me fix that up. Also, after doing the rebase, I now don't see where the TARGET_C_INT_WIDTH change came in, so I may have had an older version of rocko. Let me get that fixed up.

agherzan:
BTW. Why are we updating this in a stable branch? Wouldn't it make sense to keep it in master and release it along with the next stable release? Are there known issues with 1.21 which forces us to carry on with this update? I'm not trying to enforce maintenance policies just want to know them.

otavio:
As Rust is known to keep backward compatibility I see no reason to not upgrade it.

jcreekmore:
Furthermore, while there are no known issues with 1.21, the Rust community highly encourages you to keep on the latest stable compiler, for better or worse. Library writers tend to update their code for the latest stable compiler and, only by some loose conventions, will support back to arbitrary versions. So, to get bug fixes in some libraries, you might need to be on the latest stable even though the previous version worked with, say, 1.21.

derekstraka:
There's also no obligation for an end user to use the latest version.  We don't remove the older versions, so the PREFERRED_VERSION can be set in a local.conf to use the older version in a branch.

We have a loose guideline here: https://github.com/meta-rust/meta-rust/wiki#supported-versions

derekstraka:
@jcreekmore - Did you want to target to master first?  IIRC you mentioned that.

jcreekmore:
Uh, I did have it based on top of master (I did it right after I talked about it... the merge request is based on master now) until the SONAME branch was merged in. I can rebase back on top of master if you want me to.

derekstraka:
Ah, the title said `rocko`.  I didn't look at where it actually was targeted.

agherzan:
@otavio That is a good point but still... :)
@derekstraka this still doesn't help if something breaks on a branch that was supposed to be stable. The users, in such a case, need to switch to an older version or investigate what's going on. Doesn't sound like a nice thing for someone to encounter after updating a stable branch.
@jcreekmore Yes. You guys are the experts and if from your experience this never broke, it sounds good.

derekstraka:
I believe everything has been addressed.  I'm going to merge this in the morning just to allow any objections to come forward

hmvp:
Maybe change it to 1.24.1? https://blog.rust-lang.org/2018/03/01/Rust-1.24.1.html

jcreekmore:
Yeah, I plan on doing a PR for 1.24.1 today.

PR #189, opened 2018-02-26
libstd-rs: Set SONAME for libstd.so

agherzan:
When yocto is searching for lib dependencies it uses SONAMES and RPATHS to find
the right one. When we compile a rust package which dynamically links to rust's
libstd, yocto will not be able to find the matching dependency in shlib and
will complain with a warning. Even worse, if you have compiled go-runtime, the
build will find go-runtime as provider of libstd and will complain that you
missed a RDEPENDS to go-runtime - this is completely wrong obviously.

This patch adds SONAME to rust's stdlib. In this way yocto will create the
required provider and packages dynamically linking to it will be able to find
libstd-rs as a provider of stdlib.

Signed-off-by: Andrei Gherzan <andrei@resin.io>

### DIFF
--- a/recipes-devtools/rust/libstd-rs.inc
+++ b/recipes-devtools/rust/libstd-rs.inc
@@ -7,7 +7,7 @@ RUSTLIB_DEP = ""
 inherit cargo
 
 # Needed so cargo can find libbacktrace
-RUSTFLAGS += "-L ${STAGING_LIBDIR}"
+RUSTFLAGS += "-L ${STAGING_LIBDIR} -C link-arg=-Wl,-soname,libstd.so"
 
 S = "${RUSTSRC}/src/libstd"
 


derekstraka:
@agherzan - Are you going to want this backported as well?  I can cherry-pick as needed.  I just need to know which branchs

agherzan:
Yes. As soon as it lands in master I will push it to rocko too. As that's the one I'm mainly interested in.

derekstraka:
This one is small enough, I can just cherry-pick and push it.  I just don't like to do that for items that are more than a couple lines.

agherzan:
@derekstraka thanks. That helps.

derekstraka:
@jcreekmore - FYI.  Don't want to break you moving forward.  I doubt it will affect you

derekstraka:
@tylerwhall and @srwalter - FYI since I believe you guys use dynamic libs

derekstraka:
Cherry-picked to rocko.  Let me know if it should go into others.

agherzan:
All good. Thanks @derekstraka

tylerwhall:
:+1:

PR #192, opened 2018-05-13
Bump to Rust 1.26.2

Fulkerson:
I've updated rust to version 1.26.2.

A slight problem is `-Zremap-path-prefix-{from, to}` has stabilized and is now called `--remap-path-prefix={from, to}`. This is set in `classes/rust-common.bbclass`. Because of this this commit breaks earlier versions.

I don't know how to make the `RUST_DEBUG_REMAP` variable depend on what version of rust is selected. Would it be acceptable to drop support for older versions than 1.26.2?


derekstraka:
I'm going to wait on merging this until I create the sumo branch.  we can backport to that branch easily if there is a demand for bumping the sumo version

bachp:
From my point of view there are no objections to dropping old rust versions.

cardoe:
@bachp There's a process which this repo follows for supported Rust versions. https://github.com/meta-rust/meta-rust/wiki

Fulkerson:
We should probably wait until 1.26.1 is out and use that instead.

cardoe:
While this goes away from some of the compat guarantees we've provided in the past for people to upgrade it will be a bit messy to support that. As long as others are good with it then 👍 from me.

otavio:
I think dropping old versions is fine. Rust is good at supporting old code and this is all under 2015 edition so it is kinda safe.

In fact, if any code breaks it is likely due to a bugfix from the new stable so I think it is fine. People can always use fixed repository hashes if really need the old ones.

derekstraka:
I say we merge this once the CI completes and we can bump as required.

Fulkerson:
I've updated the pull request to 1.26.2

PR #199, opened 2018-07-04
Various fixes, Clang, TCLIBC, DEBUG_BUILD, rocko 

### LINKED ISSUE #173: The file /usr/include/llvm/LinkAllIR.h is installed by both rust-llvm-native and clang-native
## Version(s) of meta-rust
7f80e6d68b48a0a2446a606dd35901db4505d61a (HEAD -> master, upstream/master)

```
Build Configuration:
BB_VERSION        = "1.35.0"
BUILD_SYS         = "x86_64-linux"
NATIVELSBSTRING   = "universal"
TARGET_SYS        = "arm-poky-linux-gnueabi"
MACHINE           = "bananapro"
DISTRO            = "eisenhardt"
DISTRO_VERSION    = "1.0"
TUNE_FEATURES     = "arm armv7ve vfp neon vfpv4 callconvention-hard cortexa7"
TARGET_FPU        = "hard"
meta
meta-poky
meta-yocto-bsp    = "master:fcdfe09d9cc4e2977bd602dfd18f729cd9f6d73d"
meta-clang        = "master:5176136fe731f0e8e526fe7252e607dbafb3ac92"
meta-bsp_lemaker  = "development:dad114e049e7137691aa75d4c7fee77619fe75ee"
meta-xmz-mod-touch = "development:678252ce7693b4c64cdb10979e25d1e385c22e64"
meta-oe           = "master:5c9a4a08844c215ebd6f2146f50de753b8e7ecef"
meta-rust         = "master:7f80e6d68b48a0a2446a606dd35901db4505d61a"
```

## Expected result
No error

## Actual result
Error `The file /usr/include/llvm/LinkAllIR.h is installed by both rust-llvm-native and clang-native, aborting`

## Steps to reproduce
Include the `meta-rust` and `meta-clang` in your local bblayers. Then create a recipe with `DEPENDS += "clang-native"` and `bitbake` this.

I'm unsure where to report this issue. So I will here and in meta-clang https://github.com/kraj/meta-clang/issues/43

Fulkerson:
This is a few smaller fixes and tweaks I've come up with.

First is if you set DEBUG_BUILD="1" in a cargo recipe, cargo will build without --release. This mimics DEBUG_BUILD pretty well (it sets "-O0 -g" etc in normal builds). set DEBUG_BUILD_pn-whateverpackage = "1" in conf/local.conf for much shorter compile times.

Second is a change which installs rust-llvm into a subdir instead of ${prefix} as usual. It makes it possible to use meta-rust and meta-clang together. It should also fix #173.

Third adds the possibility to use TCLIBC="musl". I.e. replace glibc with musl. It seems to work and rust applications are dynamically linked to musl. Can be good to decrease the size of firmwares. It is even possible to use meta-clang, meta-rust and musl together.

Forth is adding rocko as a compatible layer. The master branch still work fine for rocko and I don't think we should artificially limit what branch you can use master on.

Fifth is target-c-int-width was incorrect on a couple of targets.

Sixth is env needs to be set to "musl" and llvm-target to "X-musl", but only for for TARGET and only when not building -native packages.

### DIFF
--- a/classes/cargo.bbclass
+++ b/classes/cargo.bbclass
@@ -23,11 +23,13 @@ B = "${S}"
 export RUST_BACKTRACE = "1"
 
 RUSTFLAGS ??= ""
-CARGO_BUILD_FLAGS = "-v --target ${HOST_SYS} --release"
+BUILD_MODE = "${@['--release', ''][d.getVar('DEBUG_BUILD') == '1']}"
+CARGO_BUILD_FLAGS = "-v --target ${HOST_SYS} ${BUILD_MODE}"
 
 # This is based on the content of CARGO_BUILD_FLAGS and generally will need to
 # change if CARGO_BUILD_FLAGS changes.
-CARGO_TARGET_SUBDIR="${HOST_SYS}/release"
+BUILD_DIR = "${@['release', 'debug'][d.getVar('DEBUG_BUILD') == '1']}"
+CARGO_TARGET_SUBDIR="${HOST_SYS}/${BUILD_DIR}"
 oe_cargo_build () {
 	export RUSTFLAGS="${RUSTFLAGS}"
 	export RUST_TARGET_PATH="${RUST_TARGET_PATH}"
--- a/classes/rust-common.bbclass
+++ b/classes/rust-common.bbclass
@@ -11,8 +11,27 @@ RUSTLIB_DEP ?= "libstd-rs"
 RUST_TARGET_PATH = "${STAGING_LIBDIR_NATIVE}/rustlib"
 RUST_PANIC_STRATEGY ?= "unwind"
 
-# Responsible for taking Yocto triples and converting it to Rust triples
+# Native builds are not effected by TCLIBC. Without this, rust-native
+# thinks it's "target" (i.e. x86_64-linux) is a musl target.
+RUST_LIBC = "${TCLIBC}"
+RUST_LIBC_class-native = "glibc"
 
+def determine_libc(d, thing):
+    '''Determine which libc something should target'''
+
+    # BUILD is never musl, TARGET may be musl or glibc,
+    # HOST could be musl, but only if a compiler is built to be run on
+    # target in which case HOST_SYS != BUILD_SYS.
+    if thing == 'TARGET':
+        libc = d.getVar('RUST_LIBC')
+    elif thing == 'BUILD' and (d.getVar('HOST_SYS') != d.getVar('BUILD_SYS')):
+        libc = d.getVar('RUST_LIBC')
+    else:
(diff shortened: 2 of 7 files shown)


classes/cargo.bbclass
```diff
@@ -23,7 +23,8 @@ B = "${S}"
 export RUST_BACKTRACE = "1"
 
 RUSTFLAGS ??= ""
-CARGO_BUILD_FLAGS = "-v --target ${HOST_SYS} --release"
+BUILD_MODE = "${@['--release', ''][d.getVar('DEBUG_BUILD') == '1']}"
```
jmesmon: Probably want to adjust `CARGO_TARGET_SUBDIR` below as well so that binaries are automatically picked up by `cargo_do_install`.

recipes-devtools/rust/rust-snapshot-1.20.0.inc
```diff
@@ -24,3 +24,4 @@ SRC_URI[rust-std-snapshot.md5sum] = "2bff47764df01c99f349908601c10478"
 SRC_URI[rust-std-snapshot.sha256sum] = "5905803e8a127f656bf253978692f0d6cf6c9206c527e4d6d7e981980618d1b6"
 SRC_URI[cargo-snapshot.md5sum] = "63aa861b029eec9f559f4fb5a10c287d"
```
jmesmon: Changing `name=` above without changing the `SRC_URI[name]` here seems like it will break 1.20.

classes/cargo.bbclass
Fulkerson: Good catch! I've updated the pull request to adjust CARGO_TARGET_SUBDIR.

recipes-devtools/rust/rust-snapshot-1.20.0.inc
Fulkerson: You are right. That file was accidentally included. I've removed it now.

Fulkerson:
It seems there are issues with the tests never completing. Is there any way to see the output?

Fulkerson:
I found another few problems.

Fifth is target-c-int-width was incorrect on a couple of targets.

Sixth is env needs to be set to "musl" and llvm-target to "X-musl", but only for for TARGET and only when not building -native packages.

derekstraka:
@Fulkerson - I recently merged in the update for rust version.  Could you rebase?  If I don't hear back by Monday EOD, I'll rebase and push it myself

Fulkerson:
@derekstraka I've rebased the fixes now.

derekstraka:
@Fulkerson - Thanks.  It's in the CI queue.  It should wrap up in a few hours and then I'm good clicking merge

PR #201, opened 2018-07-12
libstd-rs: support debug builds

srwalter:
When DEBUG_BUILD is set in conf/local.conf, libstd-rs fails to build


derekstraka:
Is this something worth adding to the CI build matrix?

srwalter:
It is a configuration we use extensively, so that would definitely be appreciated.  Are you offering, or should I look into how to do that?

derekstraka:
See #203

PR #203, opened 2018-07-12
Add more ci options for musl builds

derekstraka:


scripts/build.sh
```diff
@@ -17,3 +17,7 @@ rm -f build/conf/bblayers.conf || die "failed to nuke bblayers.conf"
 rm -f build/conf/local.conf || die "failed to nuke local.conf"
 
 ./scripts/containerize.sh bitbake ${BUILD_TARGETS} || die "failed to build"
+
+echo "DEBUG_BUILD=1" >> build/conf/local.conf
```
srwalter: This needs to use bitbake syntax, i.e. DEBUG_BUILD="1", so:

echo 'DEBUG_BUILD="1"' >> build/conf/local.conf

derekstraka:
Hmm, I'm getting the following on aarch64 when debug is enabled

```
rustc: /var/build/build/tmp/work/x86_64-linux/rust-llvm-native/1.26.2-r0/rustc-1.26.2-src/src/llvm/lib/Target/AArch64/AArch64InstructionSelector.cpp:856: virtual bool {anonymous}::AArch64InstructionSelector::select(llvm::MachineInstr&, llvm::CodeGenCoverage&) const: Assertion `SrcSize == 32 && DstTy.getSizeInBits() == 16 && "unexpected G_EXTRACT types"' failed.
```

@Fulkerson any thoughts?

Fulkerson:
@derekstraka I've never used DEBUG_BUILD globally on all packages. The resulting image would never fit on the targets I build for. Asserts are only enabled in LLVM when building debug builds so if it's not present in upstream it might have something to do with our target specification.

rustc nightly curretly says this is it's target:

`rustc --target aarch64-unknown-linux-gnu --print target-spec-json -Z unstable-options`

```json
{
  "abi-blacklist": [
    "stdcall",
    "fastcall",
    "vectorcall",
    "thiscall",
    "win64",
    "sysv64"
  ],
  "arch": "aarch64",
  "data-layout": "e-m:e-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128",
  "dynamic-linking": true,
  "env": "gnu",
  "executables": true,
  "has-elf-tls": true,
  "has-rpath": true,
  "is-builtin": true,
  "linker-flavor": "gcc",
  "linker-is-gnu": true,
  "llvm-target": "aarch64-unknown-linux-gnu",
  "max-atomic-width": 128,
  "os": "linux",
  "position-independent-executables": true,
  "pre-link-args": {
    "gcc": [
      "-Wl,--as-needed",
      "-Wl,-z,noexecstack"
    ]
  },
  "relro-level": "full",
  "target-c-int-width": "32",
  "target-endian": "little",
  "target-family": "unix",
  "target-pointer-width": "64",
  "vendor": "unknown"
}
```

We are missing some stuff (abi blacklist for example) but it doesn't look like anything major.

srwalter:
I'm able to reproduce this failure as well.  I did notice a lot of this warning in my build output:

00:58:05.100 | '+v8' is not a recognized feature for this target (ignoring feature)

Possibly related?

cardoe:
That could certainly be related to the issue @srwalter.

I added `--print target-spec-json` upstream with the goal of having it dump out the closest target and then we'd modify that target with the necessary bits because its pretty awful trying to keep the target files in sync but its never reached stable so we can't use it.

srwalter:
Closing for now.  Can be re-opened once updated for new CI

PR #204, opened 2018-07-13
libstd-rs: use a separate variable for --features

srwalter:
This will allow features to be easily modified through a bbappend file
in another layer.

The last local commit we had to meta-rust was for this purpose.  This commit would allow us to replace our local commit with a bbappend instead.


recipes-devtools/rust/libstd-rs_1.26.2.bb
```diff
@@ -5,7 +5,7 @@ LIC_FILES_CHKSUM = "file://../../COPYRIGHT;md5=99c369ad81a36cd5b27f6c6968d01055"
 
 # Don't use jemalloc as it doesn't work for many targets.
 # https://github.com/rust-lang/rust/pull/37392
-CARGO_BUILD_FLAGS += "--features 'panic-unwind'"
+CARGO_FEATURES = "panic-unwind"
```
derekstraka: Do we want this to be `?=` to provide a sane default or is this meant to be an override?

PR #205, opened 2018-07-22
rust-common: fix remap-path-prefix for 1.26

tylerwhall:
Before the flag was stabilized, it required two arguments of the form
"from=<from_path>" and "to=<to_path>" respectively. The stabilized
version uses one argument of the form "<from_path>=<to_path>".

Unfortunately the old format is still parsed successfully, but results
in attempting to replace the literal paths "from" and "to".

https://github.com/rust-lang/rust/issues/41555#issuecomment-320951103


srwalter:
This is a pretty obviously correct change.  I'll merge in 24 hours or so

PR #211, opened 2018-10-20
meta-rust: Add thud to compatible layer releases

kraj:


derekstraka:
Resolved in #210

Issue #209, opened 2018-10-07
cargo-native build failed


mark2b:
## Version(s) of meta-rust
sumo
## Version(s) of poky and/or oe-core
2.5 sumo
## Expected result
cargo-native compiled successfully 
## Actual result
cargo-native build failed because of transitive dependency of newest version zlib-sys-1.0.23
**cannot find function `read_to_string` in module `fs`**
**read_to_string** exists starting from rust 1.26, but sumo installs rust earlier versions

It seems to come from dependency **flate2-1.0** -> **zlib-sys-1.0**

## Steps to reproduce
bitbake cargo-native

```
ERROR: cargo-native-0.25.0-r0 do_compile: Function failed: do_compile (log file is located at /workspace/tmp/work/x86_64-linux/cargo-native/0.25.0-r0/temp/log.do_compile.895)
ERROR: Logfile of failure stored in: /workspace/tmp/work/x86_64-linux/cargo-native/0.25.0-r0/temp/log.do_compile.895
Log data follows:
| DEBUG: Executing shell function do_compile
| NOTE: cargo = /workspace/tmp/work/x86_64-linux/cargo-native/0.25.0-r0/cargo-0.24.0-x86_64-unknown-linux-gnu/cargo/bin/cargo
| NOTE: rustc =
| NOTE: /workspace/tmp/work/x86_64-linux/cargo-native/0.25.0-r0/cargo-0.24.0-x86_64-unknown-linux-gnu/cargo/bin/cargo build -v --target x86_64-linux --release
|     Updating registry `https://github.com/rust-lang/crates.io-index`
|  Downloading serde v1.0.79
|  Downloading filetime v0.1.15
|  Downloading tempdir v0.3.7
|  Downloading same-file v0.1.3
|  Downloading tar v0.4.17
|  Downloading serde_derive v1.0.79
|  Downloading failure v0.1.2
|  Downloading fs2 v0.4.3
|  Downloading toml v0.4.8
|  Downloading crossbeam v0.3.2
|  Downloading git2 v0.6.11
|  Downloading num_cpus v1.8.0
|  Downloading termcolor v0.3.6
|  Downloading home v0.3.3
|  Downloading jobserver v0.1.11
|  Downloading git2-curl v0.7.0
|  Downloading semver v0.8.0
|  Downloading flate2 v1.0.3
|  Downloading libgit2-sys v0.6.19
|  Downloading crypto-hash v0.3.1
|  Downloading atty v0.2.11
[...]

PR #212, opened 2018-11-05
Use libssh2 from system and upgrade cargo to 0.31.0

otavio:
This pull request enables the use of pkgconfig of libssh2 so it links against Yocto Project library and bumps the cargo utility to 0.31.0 release.

Fixes: #209.


otavio:
I updated the PR to make it use `cargo 0.30.0` for building; please review this and let me know if there is anything else need to be done for merging it.

stfl:
This needs to be backported to Sumo as well.

PR #213, opened 2018-11-06
Add support for rust 1.30.0

iohe:
My first PR, be gentle.
Adding rust 1.30.0 to meta-rust


recipes-devtools/cargo/cargo-0.30.0.inc
```diff
@@ -0,0 +1,13 @@
+CARGO_SNAPSHOT = "cargo-0.29.0-${BUILD_ARCH}-unknown-linux-gnu"
```
otavio: Please use 0.30.0 here as well.

recipes-devtools/cargo/cargo-0.30.0.inc
```diff
@@ -0,0 +1,13 @@
...
+
+SRC_URI[cargo-snapshot-x86_64.md5sum] = "f9d283378b9b0d8457f737c2a6bfbc67"
+SRC_URI[cargo-snapshot-x86_64.sha256sum] = "a5e7749767c47669ed9b6e32c6fb8eda6b0fe1c63ac73a6d4666b5c1352bad24"
+
```
otavio: Those will change.

recipes-devtools/cargo/cargo-0.30.0.inc
ashie: Probably using the previous version here is intentional.
It's for building rust/cargo themselves (stage0).

* https://github.com/rust-lang/rust/blob/master/src/stage0.txt
* https://github.com/rust-lang/rust/blob/master/src/bootstrap/README.md

```
# Currently Rust always bootstraps from the previous stable release, and in our
# train model this means that the master branch bootstraps from beta, beta
# bootstraps from current stable, and stable bootstraps from the previous stable
# release.
```

otavio: Agreed; I didn't see it.

Thanks for the reply on top of mine :+1:

recipes-devtools/rust/rust-source-1.30.0.inc
```diff
@@ -0,0 +1,22 @@
+# Specifics for Rust 1.30.0
+
+# SRC_URI += "https://static.rust-lang.org/dist/rustc-${PV}-src.tar.gz;name=rust"
+SRC_URI += " \
+        https://static.rust-lang.org/dist/rustc-${PV}-src.tar.gz;name=rust \
+        file://skip-crate-build-id-check.patch;patchdir=${WORKDIR}/rustc-${PV}-src \
```
otavio: This patch file has not been committed, please update the PR to include it or drop it.

srwalter:
We're past version 1.30 already

PR #214, opened 2018-11-10
Fix building llvm (inspirations from meta-clang)

kraj:
This make llvm cross build work better and we can compile libstd-rs, target rust still fails with SIGSEGV in rustc but that was failing always


srwalter:
Looks good to me

PR #215, opened 2018-11-12
Fix TARGET_CXXFLAGS in cargo_common

myagley:
`TARGET_CFLAGS` is used twice, copying `CXXFLAGS` into `TARGET_CFLAGS`.

I have a recipe failing to build because some C++-only flags are being appended to `CFLAGS`. It looks like this is the source of the issue. Making the change below solved the issue.


myagley:
Yeah, that's what I figured. :)

PR #226, opened 2019-01-01
Update to rust version 1.31.1

Fulkerson:
Also make use of the sources vendored in the rustc package instead of letting bitbake download the sources. This makes updating version a bit easier since there's no longer any need to extract crates using cargo-bitbake for rust itself.

"build" in config.toml has to be set to the same as the snapshot or cargo won't find any libcore etc when building the stage0 tools.


srwalter:
I'm getting a failure building rust-cross-i586 in do_rust_setup_snapshot:

    line 113: /home/swalter/build-bundle/poky/sim-mono/tmp/work/x86_64-linux/rust-cross-i586/1.31.1-r0/rust-snapshot-components/*/install.sh: No such file or directory

Fulkerson:
I believe I fixed that in the push I did a few hours ago. Basically rust-cross doesn't need the snapshot so I removed the include. But I also changed do_rust_setup_snapshot to always install the correct snapshot and avoid any others that may have been generated which broke rust-cross so that function needed to be overidden.

If you fetch the patch again I think it should work!

PR #228, opened 2019-01-14
rust: don't configure thumb if not in TUNE_FEATURES

dandedrick:
ARM_INSTRUCTION_SET, and therefore ARM_THUMB_OPT also, expect to only be
used in conjunction with the thumb TUNE_FEATURES. So thumb-mode should only
be added if thumb is in TUNE_FEATURES and ARM_THUMB_OPT is thumb.

The commit https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/meta/conf/machine/include/arm/arch-armv7a.inc?id=3e7d26e31a724009e69788460737e07163a14051
caused armv7a.inc to set ARM_INSTRUCTION_SET to thumb and the commit
message explicitly states that the change should have no effect unless
TUNE_FEATURES also includes thumb. This change to rust will allow that
assumption to continue to be true.

### DIFF
--- a/recipes-devtools/rust/rust.inc
+++ b/recipes-devtools/rust/rust.inc
@@ -78,10 +78,11 @@ def llvm_features_from_tune(d):
     if 'dsp' in feat:
         f.append("+dsp")
 
-    if d.getVar('ARM_THUMB_OPT') is "thumb":
-        if not feat.isdisjoint(v7):
-            f.append("+thumb2")
-        f.append("+thumb-mode")
+    if 'thumb' in feat:
+        if d.getVar('ARM_THUMB_OPT') is "thumb":
+            if not feat.isdisjoint(v7):
+                f.append("+thumb2")
+            f.append("+thumb-mode")
 
     if 'cortexa5' in feat:
         f.append("+a5")


recipes-devtools/rust/rust.inc
```diff
@@ -78,10 +78,11 @@ def llvm_features_from_tune(d):
     if 'dsp' in feat:
         f.append("+dsp")
 
-    if d.getVar('ARM_THUMB_OPT') is "thumb":
-        if not feat.isdisjoint(v7):
-            f.append("+thumb2")
-        f.append("+thumb-mode")
+    if 'thumb' in feat:
+        if d.getVar('ARM_THUMB_OPT') is "thumb":
```
bachp: Isn't just checking for the feature enough? Or why is this check of `ARM_THUMB_OPT` needed too?

dandedrick: Since there's not much documentation that I can find on the variables involved other than ```TUNE_FEATURES``` I'm not actually 100% certain. I think though that based on the code it's possible that ```TUNE_FEATURES``` will include thumb but ```ARM_INSTRUCTION_SET``` will be arm, in which case ```ARM_THUMB_OPT``` will be arm. I'm assuming that if someone sets ```ARM_INSTRUCTION_SET``` to arm that they actually want it to be compiled as arm even though thumb is in the ```TUNE_FEATURES```.

PR #229, opened 2019-01-16
Switchable Rust release channel

cosmo0920:
We want to switch Rust release channel because we need to try cargo's unstable cli options such as [offline mode](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/unstable.md#offline-mode).
Any advice and suggestions will be appreciated.


janderholm:
If all you want to do is try nightly options in cargo you might get by by setting the environment variable RUSTC_BOOTSTRAP=1. This is what the bootstrapping process does to make a stable compiler build a new compiler since the compiler contains unstable code.

But I think it would be nice with a patch that promotes offline mode to stable in meta-rust so that we can enable it in all builds.

cosmo0920:
> But I think it would be nice with a patch that promotes offline mode to stable in meta-rust so that we can enable it in all builds.

Hmm, you are right.
I'm closing this PR for now.
We also investigating how to enable offline mode on stable rust compiler and cargo.

PR #230, opened 2019-01-17
Update to rust version 1.32.0

janderholm:
Update to rust version 1.32.0 and build the cargo version shipped with the release.

The source of cargo used in a release is shipped with the rust source.
Use this instead of the source from github to make sure cargo and rust
are always kept in sync.

This also makes updating rust version a bit easier.


srwalter:
Looks like you have some unrelated commits on that branch

janderholm:
I don't know that they are unrelated since the update to 1.32.0 of cargo depends on the changes to what source is used when building cargo. It doesn't seem like github can handle pull requests that depend on each other but I've created two new pull requests (#231, #232) a la https://graysonkoonce.com/stacked-pull-requests-keeping-github-diffs-small/ instead.

PR #244, opened 2019-08-01
Added support for rust 1.36.0

mavax:
I'm not sure if there is any rule against skipping versions. Happy to send separate PRs for moving to 1.35 and then 1.36 if that's the case.


recipes-devtools/cargo/cargo/0001-Disable-http2.patch
```diff
@@ -19,11 +19,11 @@ index 8238380861d9..ced1defea459 100644
 --- a/Cargo.toml
 +++ b/Cargo.toml
 @@ -24,7 +24,7 @@ bytesize = "1.0"
- crates-io = { path = "src/crates-io", version = "0.23" }
```
srwalter: Won't changing this patch in place prevent older versions of cargo from building?

srwalter:
Closing in favor of #245

Issue #246, opened 2019-08-06
SSL CA cert error when cargo fetches index


janderholm:
Cargo cannot fetch crates.io index when trying to compile a package that has been checked out using `devtool modify`.

## Version(s) of meta-rust
master (c5ff05d)

## Version(s) of poky and/or oe-core
thud

## Expected result
cargo should be able to fetch index

## Actual result

```
| warning: spurious network error (2 tries remaining): [77] Problem with the SSL CA cert (path? access rights?); class=Net (12)
| warning: spurious network error (1 tries remaining): [77] Problem with the SSL CA cert (path? access rights?); class=Net (12)
```

srwalter:
Issue should be fixed by #247

PR #247, opened 2019-08-06
Set cainfo to certificates in sysroot-native

janderholm:
This solves issue #246 

When a sstate-cache is used sometimes the certificates are not available at the compile time path anymore. Typically they point to a file that existed when building another package. Possibly even another user it a shared sstate-cache is used. The required certificates are available in sysroot-native and can be selected with the cainfo configuration instead to solve the connection problems.

### DIFF
--- a/classes/cargo_common.bbclass
+++ b/classes/cargo_common.bbclass
@@ -58,6 +58,11 @@ cargo_common_do_configure () {
 		multiplexing = false
 	EOF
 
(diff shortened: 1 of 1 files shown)


mvertescher:
Ran into #246 when using a sstate-cache. The compile time CA path in libcurl was from a different user. Tested out this patch and it solved my issue.

Thanks for this fix @janderholm! Any chance we could merge this PR?

PR #257, opened 2019-12-27
rust: Use Python3 native for build

alistair23:
Use Python3 as the native Python instead of Python2.

Signed-off-by: Alistair Francis <alistair@alistair23.me>


alistair23:
Ping?

PR #259, opened 2020-01-06
Update to Rust 1.40.0

ColinFinck:
Depends on https://github.com/meta-rust/meta-rust/pull/257


mischief901:
I've tested this and it works. Any idea when it will be reviewed and merged? I'd prefer not to use a fork to use 1.40

PR #279, opened 2020-10-07
Disable LIBGIT2_SYS_USE_PKG_CONFIG due to incompatibility with libgit2 0.28.x

ColinFinck:
libgit2 0.28.x is shipped by latest Yocto Dunfell.
According to https://github.com/rust-lang/git2-rs/issues/458#issuecomment-522567539, there are no compatibility guarantees between libgit2-sys and arbitrary system libgit2 versions, so better keep this turned off.

Fixes "invalid version 3 on git_proxy_options" during build.

Further references:
* https://github.com/rust-lang/rust/issues/63476
* https://github.com/rust-lang/git2-rs/issues/458
* https://bugs.gentoo.org/707746#c1

### DIFF
--- a/recipes-devtools/cargo/cargo.inc
+++ b/recipes-devtools/cargo/cargo.inc
@@ -3,7 +3,7 @@ HOMEPAGE = "https://crates.io"
 LICENSE = "MIT | Apache-2.0"
 SECTION = "devel"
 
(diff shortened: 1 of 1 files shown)


ColinFinck:
@srwalter ?

tylerwhall (approved):
This works in my old warrior environment.

I can't find a reason for originally using the system version other than it's generally good practice.

PR #284, opened 2020-12-18
add rust 1.47

danc86:


danc86:
I actually wanted to add 1.48.0, which is the latest stable release, but Rust has made some changes to the bootstrap process since 1.47 that breaks these recipes and I haven't quite figured it out yet. I'll try again tomorrow.

srwalter (approved):

PR #285, opened 2021-01-07
Use proper llvm-target for armv7

janderholm:
arm-unknown-linux-gnueabihf was incorrectly used instead of
armv7-unknown-linux-gnueabihf when building for some Cortex A SoCs.
This may cause segfaults in non trivial rust applications on ARMv7
when e.g. +a7 is passed to LLVM. It seems to differ between different
versions of the compiler and LLVM versions.

### DIFF
--- a/recipes-devtools/rust/rust.inc
+++ b/recipes-devtools/rust/rust.inc
@@ -72,9 +72,9 @@ def llvm_features_from_tune(d):
     if 'mips32r2' in feat:
         f.append("+mips32r2")
 
-    v7=frozenset(['armv7a', 'armv7r', 'armv7m', 'armv7ve'])
-    if (not mach_overrides.isdisjoint(v7)) or (not feat.isdisjoint(v7)):
-        f.append("+v7")
+    if target_is_armv7(d):
+        f.append('+v7')
+
     if ('armv6' in mach_overrides) or ('armv6' in feat):
         f.append("+v6")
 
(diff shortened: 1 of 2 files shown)


janderholm:
The code for figuring out something is armv7 is broken out from llvm_features_from_tune and used to select the llvm target as well.

recipes-devtools/rust/rust.inc
```diff
@@ -154,6 +154,15 @@ TARGET_C_INT_WIDTH[arm] = "32"
...
 FEATURES[arm] = "+v6,+vfp2"
 
+## armv7-unknown-linux-gnueabihf
+DATA_LAYOUT[armv7] = "e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64"
+LLVM_TARGET[armv7] = "${RUST_TARGET_SYS}"
+TARGET_ENDIAN[armv7] = "little"
+TARGET_POINTER_WIDTH[armv7] = "32"
+TARGET_C_INT_WIDTH[armv7] = "32"
+MAX_ATOMIC_WIDTH[armv7] = "64"
+FEATURES[armv7] = "+v7,+thumb2"
```
srwalter, line 164: Previously this would have had +vfp2, is that implied by v7 or should it still be here?

janderholm: I've added it back and updated the pull request!

srwalter (approved):

PR #289, opened 2021-01-28
rust-llvm: Use early variable assignment for the license checksum in …

floion:
…rust-llvm.inc

Because rust-llvm-ncsa.inc requires rust-llvm.inc we do not want that
the latter overrides the license checksum set in the former.

Signed-off-by: Florin Sarbu <florin@balena.io>

### DIFF
--- a/recipes-devtools/rust/rust-llvm.inc
+++ b/recipes-devtools/rust/rust-llvm.inc
@@ -5,7 +5,7 @@ SRC_URI += "file://0002-llvm-allow-env-override-of-exe-path.patch"
 
 S = "${RUSTSRC}/src/llvm-project/llvm"
 
-LIC_FILES_CHKSUM = "file://LICENSE.TXT;md5=8a15a0759ef07f2682d2ba4b893c9afe"
+LIC_FILES_CHKSUM ?= "file://LICENSE.TXT;md5=8a15a0759ef07f2682d2ba4b893c9afe"
 
 inherit cmake python3native
 


floion:
Hi @srwalter who should I add for reviewing this? Thanks!

floion:
@ColinFinck maybe?

ColinFinck (approved):
LGTM, although I also have no push rights here. Only @srwalter can merge.

floion:
Thanks @ColinFinck

jmesmon (approved):

shr-project:
@jmesmon is this correct?

Now I see:
ERROR: QA Issue: rust-llvm-native: LIC_FILES_CHKSUM points to an invalid file: TOPDIR/BUILD/work/x86_64-linux/rust-llvm-native/1.47.0-r0/rustc-1.47.0-src/src/llvm-project/llvm/COPYRIGHT [license-checksum]

jmesmon:
Weird. This definitely built for be fine once. But it does fail that way now on rebuilds.

jmesmon:
https://github.com/meta-rust/meta-rust/pull/296 should fix this

shr-project:
> Because rust-llvm-ncsa.inc requires rust-llvm.inc we do not want that
> the latter overrides the license checksum set in the former.

So it would be enough to set LICENSE and LIC_FILES_CHKSUM _after_ require rust-llvm.inc, right?

PR #296, opened 2021-02-07
rust-source & rust-llvm: put LICENSE and LIC_FILES_CHKSUM for LLVM into source incs

jmesmon:
Rather than have a different include depending on the license, just drop
this information into our source incs and use those variables in
rust-llvm.inc.

This avoids some annoying prioritization issues, and fixes a current "2nd
time you build" issue where LIC_FILES_CHKSUM is not set properly introduced in #289


recipes-devtools/rust/rust-llvm.inc
```diff
@@ -1,12 +1,12 @@
 SUMMARY = "LLVM compiler framework (packaged with rust)"
-LICENSE ?= "Apache-2.0-with-LLVM-exception"
+
+LICENSE = "${LLVM_LICENSE}"
+LIC_FILES_CHKSUM = "${LLVM_LIC_FILES_CHKSUM}"
```
shr-project, line 4: What's the reasonding behind this?

Should rust-llvm contain both files in LIC_FILES_CHKSUM or just this LLVM_LIC_FILES_CHKSUM?

If the later then I don't understand why https://github.com/meta-rust/meta-rust/pull/289#issuecomment-774705691 changed it to weak assignment (?=), because as log as  rust-source-${PV}.inc is included before rust-llvm.inc in rust-llvm_*.bb, then normal assignments should work fine.

If both should be included then += should be used here like cargo recipes do.

Building something twice really shouldn't have any impact on how bitbake parses these assignments.

Anyway "bitbake -e rust-llvm" will always help to show how it got assigned in the end.

Now it shows
```
# $LIC_FILES_CHKSUM [3 operations]
#   set /oe-core/meta/conf/documentation.conf:265
#     [doc] "Checksums of the license text in the recipe source code."
#   set? /meta-rust/recipes-devtools/rust/rust-source-1.47.0.inc:13
#     "file://COPYRIGHT;md5=93a95682d51b4cb0a633a97046940ef0"
#   set /meta-rust/recipes-devtools/rust/rust-llvm.inc:4
#     "${LLVM_LIC_FILES_CHKSUM}"
# pre-expansion value:
#   "${LLVM_LIC_FILES_CHKSUM}"
LIC_FILES_CHKSUM="file://LICENSE.TXT;md5=8a15a0759ef07f2682d2ba4b893c9afe"
```
which confirms that the weak assignment shouldn't be needed if LLVM_LIC_FILES_CHKSUM is supposed to overwrite LIC_FILES_CHKSUM from rust-source-*.inc

jmesmon: This PR removes rust-llvm-ncsa.inc by changing how we track licenses for llvm, removing the need for some of the `?=` assignments. So we don't need to override these anymore. IOW: no more `rust-llvm-ncsa.inc` override, no more need for `?=` in this area.

(As background: the whole reason for this is that llvm changed it's licensing a while back from NCSA to Apache-2.0-with-LLVM-exception. I'm very tempted to drop support for the older rust versions at some point).

shr-project: I think this is simpler solution:
https://github.com/meta-rust/meta-rust/pull/297

PR #297, opened 2021-02-07
Fix https://github.com/meta-rust/meta-rust/pull/289

shr-project:


jmesmon (approved):
Yep, this does work and fixes the immediate issue. Still probably want to consider some reorganization to how we're handling this (like #296)

It does appear that our older rust-llvm targets (all the ncsa ones) are currently unbuildable anyhow (due to some new targets we've added)

PR #305, opened 2021-02-09
rust: Drop v8.1a tune for aarch64

akiernan:
Building for "+v8.1a"  breaks on (at least) Cortex A53 (Armv8-A) on startup with `Illegal instruction`.

Fixes: 26609f46d93c ("rust.inc: use 'v8.1a' feature when building for aarch64 instead of 'v8'")

### DIFF
--- a/recipes-devtools/rust/rust.inc
+++ b/recipes-devtools/rust/rust.inc
@@ -66,9 +66,6 @@ def llvm_features_from_tune(d):
     if 'neon' in feat:
         f.append("+neon")
 
-    if 'aarch64' in feat:
-        f.append("+v8.1a")
-
     if 'mips32' in feat:
         f.append("+mips32")
 


jmesmon (approved):

kraj:
while I am ok with using common minimum, perhaps it could be made something user can set via a pckageconfig or something, since performance might be improved by using appropriate -march

jmesmon:
Agreed, perhaps there is some extra `TUNE_FEATURES` values or other existing variable that we could use?

kraj:
yeah presence of `armv8-2a` in `TUNE_FEATURES` and/orr `crypto` e.g. can be used to construct this logic

shr-project:
Maybe appropriate cpu setting (instead of generic used now) would help as well.

Issue #309, opened 2021-02-15
The cargo class does not factor in `CARGO_SRC_DIR`


BeneSim:
## Version(s) of meta-rust
master c4bef774e8361bbd613e9e7ce00cd9c37f61d513

## Version(s) of poky and/or oe-core
poky: -
oe-core: dunfell

## Expected result
When using `cargo-bitbake` within a package of a Cargo workspace the recipe should take care that only this package will get build and deployed.

## Actual result
Even though `cargo-bitbake` populates the `CARGO_SRC_DIR` variable with the correct relative path to the target package the whole Cargo workspace gets build and deployed. This is due to the fact that `CARGO_SRC_DIR` does not get factored in by the `cargo.bbclass` when setting the default value for `MANIFEST_PATH` https://github.com/meta-rust/meta-rust/blob/341a3c77bbbde5b321471f193d31e317e96a9ef3/classes/cargo.bbclass#L26

## Steps to reproduce
Crate a Cargo workspace with a at least two binary subpackages, run `cargo-bitbake` and use the recipe within your openembedded project.

## Solution
Actually there are two viable solutions in my opionion

### 1.) Factor in `MANIFEST_PATH` in the `cargo.bbclass`.
This would be quite simple, just change
```
MANIFEST_PATH ??= "${S}/Cargo.toml"
```
to
```
MANIFEST_PATH ??= "${S}/${CARGO_SRC_DIR}/Cargo.toml"
```
However this could break with existing code, i.e. when `CARGO_SRC_DIR` is set to a wrong path or `S` gets set to a different value.

### 2.) Set `MANIFEST_PATH` with `cargo-bitbake`
We could simply set `MANIFEST_PATH` within the `bitbake.template` 
```
...
SRCREV = "{project_src_rev}"
S = "${{WORKDIR}}/git"
CARGO_SRC_DIR = "{project_rel_dir}"
MANIFEST_PATH = "${{S}}/${{CARGO_SRC_DIR}}/Cargo.toml"
{git_srcpv}
...
```
we could also get rid of `CARGO_SRC_DIR` altogether in the above example.

I'll PR both solutions so you can chose, or maybe there is a better way to tackle this :)

BeneSim:
Closing due to merge of https://github.com/meta-rust/meta-rust/pull/310

PR #310, opened 2021-02-15
Factor in `CARGO_SRC_DIR` when setting `MANIFEST_PATH`.

BeneSim:
Fixes Cargo workspace builds, see #309 . There is an alternative PR up within `cargo-bitbake` https://github.com/meta-rust/cargo-bitbake/pull/23.


jmesmon (approved):
Adding this here rather than cargo-bitbake seems fine.

PR #312, opened 2021-02-16
rust-native shouldn't depend on TARGET variables

janderholm:
Similar to 3b783652cc9e0e0ff95fdf151eab3e178b5f1667, this makes sure
rust-native does not depend on target variables. This allow one
rust-native to be shared between machines of different architectures
again.

For the record, the following guide was used to find out why
do_rust_gen_targets differed between different machines.
https://wiki.yoctoproject.org/wiki/TipsAndTricks/Understanding_what_changed_(diffsigs_etc)


danc86 (approved):

jmesmon:
(@danc86: feel free to merge these in the future if they look good to you)

PR #313, opened 2021-02-16
nativesdk: Implement cross canadian support

### LINKED ISSUE #236: nativesdk support
## Version(s) of meta-rust
master branch, 0621067c560f6e3e1a4c4c08ac173393bbafad5c
poky 2.6.1, thud branch
## Expected result
rust is in the SDK and can be used to compile rust applications with the standard sdk, built with bitbake <image> -c populate_sdk
## Actual result
meta-rust does not have support for nativesdk

weltling:
This addition closes #236 requesting the SDK support. The implementation
consists on a yet minimum set of worky functionality;

- Includes rustc, rustdoc and cargo.
- Includes libstd and accompanying lib archives for host and for target.
- Integrates with the standard environment setup script.
- Integrates configurations automatically build for target.
- Tested with Linux based SDK.

The supported build host is still AMD64 only.

Note that there's no need on introducing crosssdk package as the
official snapshot package serves same way as a non SDK build.

A necessary step had to be done was to separate out some common part from
the rust recipe, so it can be shared with the cargo build.

Possible future directions:

- Offline build:
	- Support pre-cached set of crates that can be vendored.
	- Support pre-filled cargo registry.
- Support further tools like cargo-fmt, cargo-clippy, GDB support and so
on.


rwmacleod:
It looks okay to me. I don't work on the SDK much but I reviewed the patches and built packagegroup-rust-cross-canadian-qemux86-64. I'm trying to figure out how to add it to the standard or extensible SDK. Can you explain how you tested this?
Also other than the comments in the code, is there anything you think is worrisome or that you'd like to draw reviewers attention to? 

In addition I have these minor comments and/or questions:
1. Should this be in some other rust/cago/??? class:
```
+do_compile_prepend () {
+       PKG_CONFIG_PATH="${RECIPE_SYSROOT_NATIVE}/usr/lib/pkgconfig:${PKG_CONFIG_PATH}"
+}
```
2. There are a couple of trailing spaces in packagegroup-rust-cross-canadian.bb
3. There some whitespace inconsistencies in rust-cross-canadian.inc and maybe elsewhere.

Thanks and I'll test and review more once I can confirm that it not only builds but can populate and sdk and be used on another system. :)

weltling:
Thanks a lot for the review.

Regarding SDK inclusion, I've added the below to the `local.conf`:

```
TOOLCHAIN_HOST_TASK_append = " \ 
    packagegroup-rust-cross-canadian-${MACHINE} \
"
```

Then building with:

`bitbake core-image-minimal -c populate_sdk`

That will produce the SDK with the rust environment included.


I tested on build=AMD64, host=AMD64 and targets=AARCH64 under Ubuntu 18.04. 

For the tests, one can go the following way

- Unpack the resulting shar and setup the env.
  - The env setup scripts can be reviewed (though they're generated)
     sysroots/x86_64-pokysdk-linux/environment-setup.d/rust.sh
     sysroots/x86_64-pokysdk-linux/environment-setup.d/cargo.sh
  - The default target is set under sysroots/aarch64-poky-linux/home/cargo/config.
- Create some project:
```bash
$ cargo new hello --bin
$ cd hello
$ cargo build
$ file ./target/aarch64-poky-linux/debug/hello
./target/aarch64-poky-linux/debug/hello: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=d1d333e0e71027353d2bddb66bed2d0973a3fcf6, for GNU/Linux 3.14.0, with debug_info, not stripped
# copy to some host where it can run
```
- Test some more complex build:
```bash
$ git clone https://github.com/cloud-hypervisor/micro-http.git
$ cd micro-http/
$ cargo build -v
   Compiling bitflags v1.2.1
   Compiling libc v0.2.66
............................
$ readelf -h target/aarch64-poky-linux/debug/libmicro_http.rlib | grep 'Class\|File\|Machine'
File: target/aarch64-poky-linux/debug/libmicro_http.rlib(micro_http-1b597fd7448eba7d.13y5x0h0sr4cq9he.rcgu.o)
  Class:                             ELF64
  Machine:                           AArch64
File: target/aarch64-poky-linux/debug/libmicro_http.rlib(micro_http-1b597fd7448eba7d.16oi4yi23xcirlqc.rcgu.o)
  Class:                             ELF64
  Machine:                           AArch64
........................
```
Note that this one also builds the build scripts. That was a tricky point when developing and that's why the SDK requires both host and target libstd and libs. The catchword here is `build.rs` in the doc.

Another point to draw attention is probably is what i mentioned earlier about ./sysroots/aarch64-poky-linux/home/cargo/config which implies `CARGO_HOME` is set to `./sysroots/aarch64-poky-linux/home/cargo`. That can be set to something else of course, just my point was that the dependencies will probably be in the target architecture. This might or might not be touched later when it possibly comes to improvements to the offline build.

One more point is with the build target names as rust would get them. That is a tricky business as for example `aarch64-poky-linux` as a custom target requires quite some decorative work to be done around. But also, Rust looking at many points to hardcoded paths is often an issue that conditions some business that otherwise would not be necessary. That's probably to mention when one sees something going on with paths that might look tricky.

Other that that - I seed the functionality overall in a robust alpha state, there will be sure some bugs and some refactoring to do in the coming iterations.

Regarding 1. - there seem to be several tricky cases like this. Currently there's no `pkgconfig` handling in any of cargo or rust bbclasses and this case is specific to the SDK build only. The certain difference here is that when `cargo.bbclass` is used without the SDK, it's most about `RECIPE_SYSROOT`. The need on this is based on the Rust requirement to pull OpenSSL and the availability is checked using pkg-config. I saw that the SDK build infrastructure should actually serve it

http://git.yoctoproject.org/clean/cgit.cgi/poky/tree/meta/classes/nativesdk.bbclass#n71

and i indeed adedd `nativesdk-openssl` to populate that requirement, but it seemed that it gets pulled somewhere else or maybe i did it a wrong way. I'm going to verify this part once more. Regarding the `cargo.bbclass` and `pkg-config` - I'd guess just `inherit pkgconfig` would be fine for a non-SDK build, say if some crate requires pkg-config.

For the WS things - gonna grep through and fix that.

thanks!

rwmacleod:
Thanks. Here's my update before I chill out for the evening. I added those lines to the local.conf ( and saw how I should have figured that out myself!) 
$ bitbake -c populate_sdk core-image-minimal
    worked fine after I cleaned up. In a separate shell:
$ . ./environment-setup-core2-64-oe-linux 
works fine and now our rustc/cargo are from the sdk:
$ which cargo
/.../b/tsdk/sysroots/x86_64-oesdk-linux/usr/bin/cargo
But...
$ cargo --help
cargo: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /.../b/tsdk/sysroots/x86_64-oesdk-linux/usr/bin/../lib/libcrypto.so.1.1)
cargo: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /.../b/tsdk/sysroots/x86_64-oesdk-linux/usr/bin/../lib/libcurl.so.4)

$ ldd sysroots/x86_64-oesdk-linux/usr/bin/../lib/libcrypto.so.1.1
sysroots/x86_64-oesdk-linux/usr/bin/../lib/libcrypto.so.1.1: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by sysroots/x86_64-oesdk-linux/usr/bin/../lib/libcrypto.so.1.1)
        linux-vdso.so.1 (0x00007ffe02565000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fcedf2a0000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fcedf081000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fcedec90000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fcedf749000)

I'm using Ubuntu-18.04.3 which has glibc-2.27 of course.
It's likely something simple but I'm done for the day.

weltling:
Thanks for ther further check. Could you please tell

- Which Poky branch did you use? On my side it was in most case Dunfell and gatesgarth.
- Did you use uninative? If yes, it would make sense to check disabling uninative.
- Is the SDK unpackade on the same machine? 

On my side it's Ubuntu 18.04.5. On the artifacts from the last buid i see right now:

```
$ which cargo
/tmp/rust-sdk-deploy-16/sysroots/x86_64-pokysdk-linux/usr/bin/cargo
$ ldd /tmp/rust-sdk-deploy-16/sysroots/x86_64-pokysdk-linux/usr/bin/cargo
	linux-vdso.so.1 (0x00007ffe885ee000)
	libssl.so.1.1 => /tmp/rust-sdk-deploy-16/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libssl.so.1.1 (0x00007f3ae7976000)
	libcrypto.so.1.1 => /tmp/rust-sdk-deploy-16/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1 (0x00007f3ae67d1000)
	libz.so.1 => /tmp/rust-sdk-deploy-16/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libz.so.1 (0x00007f3ae795c000)
	libcurl.so.4 => /tmp/rust-sdk-deploy-16/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcurl.so.4 (0x00007f3ae78f7000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f3ae65b9000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f3ae63b1000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3ae6192000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3ae5df4000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f3ae5bf0000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3ae57ff000)
	/tmp/rust-sdk-deploy-16/sysroots/x86_64-pokysdk-linux/lib/ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2 (0x00007f3ae77df000)

```

Perhaps it's the SDK target x86_64 the issue. I'm going for that and check. Another suspiction i have is a uninative issue - i've met similar issues in the past and otherwise saw other having sometimes. I'll continue using Dunfell as the base to also exclude any instabilty in the result, lets see.

Thanks

weltling:
I've rebuilt for x86-64 as a target, but seems fine. Still Ubuntu 18.04.5 and Dunfell as a base.

```bash
$ . environment-setup-core2-64-poky-linux
$ which cargo
/tmp/rust-sdk-deploy-17/sysroots/x86_64-pokysdk-linux/usr/bin/cargo
$ ldd $(which cargo)
        linux-vdso.so.1 (0x00007ffc8d9f8000)
        libssl.so.1.1 => /tmp/rust-sdk-deploy-17/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libssl.so.1.1 (0x00007f7c2ba58000)
        libcrypto.so.1.1 => /tmp/rust-sdk-deploy-17/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1 (0x00007f7c2a8b3000)
        libz.so.1 => /tmp/rust-sdk-deploy-17/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libz.so.1 (0x00007f7c2ba3e000)
        libcurl.so.4 => /tmp/rust-sdk-deploy-17/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcurl.so.4 (0x00007f7c2b9d9000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f7c2a69b000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f7c2a493000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f7c2a274000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f7c29ed6000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f7c29cd2000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f7c298e1000)
        /tmp/rust-sdk-deploy-17/sysroots/x86_64-pokysdk-linux/lib/ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2 (0x00007f7c2b8c1000)
```

In that build also, the glibc versioned symbols imported by the particular DSO:

```bash
$ objdump -T /tmp/rust-sdk-deploy-17/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1 | grep GLIBC_ | sed 's,  w,   ,g' | cut -d' ' -f10 | sort -u
GLIBC_2.14
GLIBC_2.16
GLIBC_2.17
GLIBC_2.2.5
GLIBC_2.25
GLIBC_2.3
GLIBC_2.3.2
GLIBC_2.3.4
GLIBC_2.7
```

Whereby i don't think the symbol versions do really matter, it's just about the symbol compatibiliity between a particular Yocto release and the particular host glibc, so it's probably about the supported host vs. Poky branch.

Thanks.

rwmacleod:
1. Which Poky branch did you use? oe-core/master on top of my patches moving meta-rust to oe-core. As I'm investigating today, I may try an older branch and/or poky+meta-rust but the latter combo shouldn't be required as I haven't changed libssl.
    2. Did you use uninative? No, I didn't use uninative.
    3. Is the SDK installed on the same machine? Yes it is on the same machine.

weltling:
Ok, so that might be due to the specific build. I was further digging today, but no reproduce so far. My current local.conf is this:

```
MACHINE ??= "qemux86-64"
DISTRO ?= "poky"
PACKAGE_CLASSES ?= "package_rpm"
SDKMACHINE ?= "x86_64"
EXTRA_IMAGE_FEATURES ?= "debug-tweaks"
USER_CLASSES ?= "buildstats image-mklibs image-prelink"
PATCHRESOLVE = "noop"
BB_DISKMON_DIRS ??= "\
    STOPTASKS,${TMPDIR},1G,100K \
    STOPTASKS,${DL_DIR},1G,100K \
    STOPTASKS,${SSTATE_DIR},1G,100K \
    STOPTASKS,/tmp,100M,100K \
    ABORT,${TMPDIR},100M,1K \
    ABORT,${DL_DIR},100M,1K \
    ABORT,${SSTATE_DIR},100M,1K \
    ABORT,/tmp,10M,1K"
PACKAGECONFIG_append_pn-qemu-system-native = " sdl"
CONF_VERSION = "1"
TOOLCHAIN_HOST_TASK_append = " \
    packagegroup-rust-cross-canadian-${MACHINE} \
    packagegroup-go-cross-canadian-${MACHINE} \
"
PREFERRED_PROVIDER_virtual/kernel = "linux-dummy"
BASECANADIANEXTRAOS = ""
```

Or change the machine to `qemuarm64`. I'll go for another run to use poky master yet.

It might be useful you to share some more detail on your side then, so i could go for the same. Depending on where you are at the end of day  :)

Thanks!

rwmacleod:
I had other things to take care of today so I haven't made any progress.
My meta-rust -> oe-core patchset is currently based on:
5b509fe1c5 (origin/master, origin/HEAD, master) maintainers: Update email address for Victor
My config is:
MACHINE ??= "qemux86-64"
PACKAGE_CLASSES ?= "package_ipk"
EXTRA_IMAGE_FEATURES ?= "debug-tweaks"
USER_CLASSES ?= "buildstats image-mklibs image-prelink"
PATCHRESOLVE = "noop"
BB_DISKMON_DIRS ??= "\
    STOPTASKS,${TMPDIR},1G,100K \
    STOPTASKS,${DL_DIR},1G,100K \
    STOPTASKS,${SSTATE_DIR},1G,100K \
    STOPTASKS,/tmp,100M,100K \
    ABORT,${TMPDIR},100M,1K \
    ABORT,${DL_DIR},100M,1K \
    ABORT,${SSTATE_DIR},100M,1K \
    ABORT,/tmp,10M,1K"
PACKAGECONFIG_append_pn-qemu-system-native = " sdl"
PACKAGECONFIG_append_pn-nativesdk-qemu = " sdl"
CONF_VERSION = "1"
IMAGE_INSTALL_append = " rust-hello-world"
TOOLCHAIN_HOST_TASK_append = " \
    packagegroup-rust-cross-canadian-${MACHINE} \
"

weltling:
Thanks for sharing the conf. I compiled first using vanilla master of poky and meta-openembedded and can repro something similar:

```bash
$ ldd $(which cargo)
/tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/cargo: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1)
/tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/cargo: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcurl.so.4)
        linux-vdso.so.1 (0x00007ffc8e55e000)
        libssl.so.1.1 => /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libssl.so.1.1 (0x00007f37532fa000)
        libcrypto.so.1.1 => /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1 (0x00007f375216a000)
        libz.so.1 => /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libz.so.1 (0x00007f37532e0000)
        libcurl.so.4 => /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcurl.so.4 (0x00007f375326b000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f3751f52000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f3751d4a000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3751b2b000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f375178d000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f3751589000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3751198000)
        /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/lib/ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2 (0x00007f3753162000)
```

But on the other hand, it shows also similar with GCC

```bash
$ ldd $(which x86_64-poky-linux-gcc)
/tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/x86_64-poky-linux/x86_64-poky-linux-gcc: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/x86_64-poky-linux/x86_64-poky-linux-gcc)
        linux-vdso.so.1 (0x00007ffe1d174000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f93c702e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f93c6c3d000)
        /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/lib/ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2 (0x00007f93c73cc000)
```

Also, see what exactly it is:

```bash
$ objdump -T /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1 | grep GLIBC_2.33
objdump: warning: /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010001
objdump: warning: /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcrypto.so.1.1: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010002
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.33  stat
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.33  fstat
$ objdump -T /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/x86_64-poky-linux/x86_64-poky-linux-gcc | grep GLIBC_2.33
objdump: warning: /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/x86_64-poky-linux/x86_64-poky-linux-gcc: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0008002
objdump: warning: /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/x86_64-poky-linux/x86_64-poky-linux-gcc: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010001
objdump: warning: /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/x86_64-poky-linux/x86_64-poky-linux-gcc: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010002
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.33  lstat
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.33  stat
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.33  fstat
$ objdump -T /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcurl.so.4 | grep GLIBC_2.33
objdump: warning: /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcurl.so.4: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010001
objdump: warning: /tmp/rust-sdk-deploy-18/sysroots/x86_64-pokysdk-linux/usr/bin/../lib/libcurl.so.4: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010002
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.33  fstat
0000000000000000      DF *UND*  0000000000000000  GLIBC_2.33  stat
```

It's glibc 2.33 in master, but it's also 2.31 in dunfell. At this point i'm not sure, but i'd tend to say master is not suitable as seems it's a bug there - as similar is to see with other artifacts, like useradd, cmake, etc. I'm gonna dig more on that. If there's someone with a better expertise on the SDK front, any help would be great.

To unblock your further review, i'd suggest to do the factual test on the base of dunfell first - if that's ok on your side, one can move forward to check upper branches.

Thanks

rwmacleod:
Sure. I"m building oe-core+meta-rust+meta-oe + your sdk patch based on gatesgarth. Stay tuned.
Build Configuration:
BB_VERSION           = "1.48.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "ubuntu-18.04"
TARGET_SYS           = "x86_64-oe-linux"
MACHINE              = "qemux86-64"
DISTRO               = "nodistro"
DISTRO_VERSION       = "nodistro.0"
TUNE_FEATURES        = "m64 core2"
TARGET_FPU           = ""
meta                 = "HEAD:d3114ca369792201a5316c1ede29eb72f0868d35"
meta-rust.git        = "rsdk:03060181a11d954d538c1fb6c2fa3d9b71a68e0a"
meta-oe
meta-python
meta-networking      = "gatesgarth:cef93b7b00e620d90a610112ee574fa60b691cf8"

rwmacleod:
building oe-core+meta-rust+meta-oe + your sdk patches based on gatesgarth worked just fine:
```
$ which rustc
/.../sysroots/x86_64-oesdk-linux/usr/bin/rustc
$ ldd `which rustc`
        linux-vdso.so.1 (0x00007ffc05bfa000)
        librustc_driver-7571c27e0d047d0f.so => /.../sysroots/x86_64-oesdk-linux/usr/bin/../lib/librustc_driver-7571c27e0d047d0f.so (0x00007fd5792f3000)
        libstd-86e31077140c3cde.so => /.../sysroots/x86_64-oesdk-linux/usr/bin/../lib/libstd-86e31077140c3cde.so (0x00007fd578f4a000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd578b59000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fd578951000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fd57874d000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fd57852e000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fd578190000)
        libstdc++.so.6 => /.../sysroots/x86_64-oesdk-linux/usr/bin/../lib/../lib/libstdc++.so.6 (0x00007fd58145a000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fd577f78000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fd581416000)
$ rustc --help| head -3
Usage: rustc [OPTIONS] INPUT

Options:
```
Let's merge and we'll work out the problem related to glibc-2.33 afterward since it's not related to these two patches.
Thanks for you patience.

rwmacleod (approved):
Works well on gatesgarth but not on master due to oe-core problems.
Minor whitespace changes can be fixed later. I think we should merge this PR.

weltling:
@rwmacleod thanks for the review trying out, I rebased to the latest to fix the merge.

With the WS issues - I fixed some a day ago. Some places with bash heredoc have to stay however, as long as scripts are generated that way. 

I went to the core ML with the glibc issue in master, will continue there.

Thanks!

weltling:
As explained in the core ML conversation, the observations on master are rather expected.

https://lists.openembedded.org/g/openembedded-core/topic/80874524#148565

Thanks.

recipes-devtools/rust/rust-common.inc
```diff
@@ -0,0 +1,337 @@
...
+
+def llvm_features(d):
+    return ','.join(llvm_features_from_tune(d) +
+                    llvm_features_from_cc_arch(d) +
+                    llvm_features_from_target_fpu(d))
+
+
+## arm-unknown-linux-gnueabihf
+DATA_LAYOUT[arm] = "e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64"
+LLVM_TARGET[arm] = "${RUST_TARGET_SYS}"
```
lorenzph, line 113: Do you (or anybody else) know why `RUST_TARGET_SYS` is used for all targets?

It looks like you hit the same issue I did (namely the cross-canadian variant producing binaries for `TARGET_SYS`) and fixed that by overriding the target in the `cross-canadian.inc` files.

In my worktree I fixed this by changing `rust_gen_target` to choose the correct value based on the content of `RUST_thing_SYS` instead.

recipes-devtools/rust/rust-cross-canadian.inc
```diff
@@ -0,0 +1,224 @@
...
+
+    rustc = d.expand("${WORKDIR}/rust-snapshot/bin/rustc")
+    config.set("build", "rustc", e(rustc))
+
+    cargo = d.expand("${WORKDIR}/rust-snapshot/bin/cargo")
+    config.set("build", "cargo", e(cargo))
+
+    config.set("build", "vendor", e(True))
+
+    targets = [d.getVar("TARGET_SYS", True), "{}-unknown-linux-gnu".format(d.getVar("HOST_ARCH", True))]
```
lorenzph, line 138: Shouldn't this be `SDK_SYS` rather than `{}-unknown-linux-gnu`?

recipes-devtools/rust/rust-cross-canadian.inc
```diff
@@ -0,0 +1,224 @@
...
+    config.set("build", "target", e(targets))
+
+    hosts = ["{}-unknown-linux-gnu".format(d.getVar("HOST_ARCH", True))]
```
lorenzph, line 141: Dito

recipes-devtools/rust/rust-cross-canadian.inc
```diff
@@ -0,0 +1,224 @@
...
+
+    install -d "${SYS_BINDIR}"
+    cp build/${SNAPSHOT_BUILD_SYS}/stage2/bin/* ${SYS_BINDIR}
+    for i in ${SYS_BINDIR}/*; do
+	chrpath -r "\$ORIGIN/../lib" ${i}
+    done
+
+    install -d "${D}${libdir}"
+    cp -pRd build/${SNAPSHOT_BUILD_SYS}/stage2/lib/${TARGET_SYS}/*.so ${SYS_LIBDIR}
+    cp -pRd build/${SNAPSHOT_BUILD_SYS}/stage2/lib/${TARGET_SYS}/rustlib ${RUSTLIB_DIR}
```
lorenzph, line 182: The target build seems to prefer `libstd-rs` to provide the standard library (I don't know the exact rationale behind this but it may be related to difficulties compiling for multiple target architectures in a `native` recipe).

Would it maybe make sense to use the same approach for the SDK (i.e. a nativesdk-libstd-rs package)?

recipes-devtools/rust/rust-cross-canadian.inc
```diff
@@ -0,0 +1,224 @@
+
```
lorenzph, line 1: I'm hitting some QA warnings during packaging:

```
WARNING: QA Issue: rust-cross-canadian-aarch64: found library in wrong location: /opt/sdk/sysroots/x86_64-pokysdk-linux/usr/lib/librustc_macros-d9145fa255675431.so
rust-cross-canadian-aarch64: found library in wrong location: /opt/sdk/sysroots/x86_64-pokysdk-linux/usr/lib/libtracing_attributes-301db28eaa169a2a.so
rust-cross-canadian-aarch64: found library in wrong location: /opt/sdk/sysroots/x86_64-pokysdk-linux/usr/lib/libstd-86e31077140c3cde.so
rust-cross-canadian-aarch64: found library in wrong location: /opt/sdk/sysroots/x86_64-pokysdk-linux/usr/lib/libchalk_derive-f58cadaeafa0e1a3.so
rust-cross-canadian-aarch64: found library in wrong location: /opt/sdk/sysroots/x86_64-pokysdk-linux/usr/lib/libtest-cd5cb2eef27bf3af.so
rust-cross-canadian-aarch64: found library in wrong location: /opt/sdk/sysroots/x86_64-pokysdk-linux/usr/lib/librustc_driver-7571c27e0d047d0f.so [libdir]
```

lorenzph (commented):
I started working on packaging Rust for the SDK on and off a couple of weeks ago and just noticed that you were doing the same :-)

I still don't have a fully working state so I believe it makes sense to collaborate here.

recipes-devtools/cargo/cargo-cross-canadian_1.49.0.bb
```diff
@@ -0,0 +1,6 @@
+require recipes-devtools/rust/rust-source-${PV}.inc
+require recipes-devtools/rust/rust-snapshot-${PV}.inc
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/cargo-${PV}:"
+
+require cargo-cross-canadian.inc
```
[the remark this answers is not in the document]
weltling: Thanks for joining on this work! I'd wonder if there's more people to get this feature as the issue has been filed a couple of years ago already.

To the actual issue - could you please share which branch you base on and which layers are included. The description should probably mention that it's currently about dunfell and upwards.  If you've already some fix in mind, please PR against the feature branch or just send a patch. Otherwise please let me know how it can be reproduced, maybe that's the base poky branch or something in the config that i currently don't have.

Regrading the question about the canadian cross build - the produced binaries are dedicated to the prospective host architecture.  It is not agnostic. As meta-rust only supports amd64 as a build host, it might make things simpler, but still keeping up with teh concept is good. For the same reason, no crosssdk builds are necessary.

Thanks

recipes-devtools/rust/rust-cross-canadian.inc
weltling: Nope, here and above it is about the the host variant of libstd. I tried using a custom target there but there was issues. I suspect it's about things being hardcoded in `rustc` so using the host libstd as `rustc` wants it is the simplest here. The name for the target libstdc could be customized however.

Thanks.

recipes-devtools/rust/rust-cross-canadian.inc
weltling: It would make things more complicated as it needs libstd for the host and for the target, both are produced within the `rust-cross-canadian-`.  Find that way for now at least, as for me.

Thansk.

recipes-devtools/rust/rust-cross-canadian.inc
weltling: Lets see if i can repro it, would need more data from your setup as per the request your first comment.

Thanks.

weltling: Though please note - if your checks was on something lower than dunfell, please used dunfell first and lets move from there.

Thanks.

recipes-devtools/cargo/cargo-cross-canadian_1.49.0.bb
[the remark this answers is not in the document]
lorenzph: > Thanks for joining on this work! I'd wonder if there's more people to get this feature as the issue has been filed a couple of years ago already.

I was also surprised to see that someone else started working on this simultaneously (in particular because I checked the open PRs before starting this off).

> To the actual issue with the header - could you please share which branch you base on and which layers are included. The description should probably mention that it's currently about dunfell and upwards. If you've already some fix in mind, please PR against the feature branch or just send a patch. Otherwise please let me know how it can be reproduced, maybe that's the base poky branch or something in the config that i currently don't have.

I'm testing this on dunfell (although I'd probably end up using this on earlier versions as well). At the end of the day I suspect that this boils down to the missing dependency on `nativesdk-libssh2` (although in that case I'm also having trouble understanding why it compiles for everybody else). I don't have a build dir ready to retrieve logs but I'll revisit this later.

> Regrading the question about the canadian cross build - the produced binaries are dedicated to the prospective host architecture. It is not agnostic. As meta-rust only supports amd64 as a build host, it might make things simpler, but still keeping up with teh concept is good. For the same reason, no crosssdk builds are necessary.

I'm having a hard to time to find the correct Yocto terminology for this as (in contrast to GCC) a rustc compiled for `SDK_SYS` seems to be perfectly capable to compile for any other architecture which is supported by its LLVM and for which a `std` build is available (and I'd expect cargo to be even more flexible here).

When starting off on this quest I even attempted to configure `rust-native` in such a way that it would produce binaries for `BUILD_SYS` / `SDK_SYS` and the target's architecture (however these attempts failed due to various assumptions in the Yocto toolchain which prevented me from installing the `cross` and `crosssdk` GCC variants).

So far I ended up packaging the compiler parts in a `nativesdk-rust` package and the `std` parts in a `nativesdk-libstd-rs` package. In this setup the `cross-canadian-rust` package only provides a target configuration file (similar to the approach chosen for the current `cross` recipe) and a linker wrapper script (as at least in my setup the linker is not able to find its support libraries without explicitly passing it via `--sysroot`).

recipes-devtools/rust/rust-cross-canadian.inc
lorenzph: But isn't `SDK_SYS` the host of the binaries produced by the cross-canadian recipe? I've managed to compile for the `SDK_SYS` host (although when producing a complete compiler via the cross-canadian bbclass dedicated compiler wrappers for `HOST_SYS` / `SDK_SYS` were needed).

recipes-devtools/rust/rust-cross-canadian.inc
lorenzph: Yes, that's a big advantage of having the sysroot shipped as part of the cross-compilation toolchain. However, this also means that users of the SDK will not link against the same set of `std` libraries shipped in the target root filesystem (I am not sure if this can cause any problems as the versions should be identical).

recipes-devtools/rust/rust-cross-canadian.inc
weltling: Hi,

the existing mechanism in `meta-rust` will generate a definition for the custom target. What it is about here is that `x86_64-unknown-linux-gnu` is used ofr the host libstd, because it's a built in target in Rust. The end target is custom and is served by the json file, for example `aarch64-poky-linux.json`. Custom targets seem to be a tricky business as Rust has several things hardcoded.

Thanks.

recipes-devtools/rust/rust-cross-canadian.inc
weltling: That's a good point. I wouldn't see this as big, as in most case the ABI should be same. Say if later the binary artifacts are to be included into an image, the Poky QA task will bail out if it found some issues. It probably would need more thought, as this will need the `libstd` recipe to be integrated with the cross canadian mechanism.

Thanks.

recipes-devtools/cargo/cargo-cross-canadian_1.49.0.bb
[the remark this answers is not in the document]
weltling: Yes, the Rust builds stage2 are in general capable of cross compiling. What they need is:

- `rustc`
- libstd for host
- libstd for target

If desired, a custom target json can be used, too.

In case host and target are same, obviously one can get away just with one version of `libstd`.

Thanks.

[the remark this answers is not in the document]
lorenzph: So I've managed to fix the compilation issue by manually adding `nativesdk-libssh2` to `DEPENDS`. I am doing all my Yocto builds in a minimal Debian 10 container so I am wondering if you maybe didn't see this because of host contamination.

After that I was hit with a packaging QA warning complaining about a missing provider for libcurl

`ERROR: cargo-cross-canadian-aarch64-1.49.0-r0 do_package_qa: QA Issue: /opt/sdk/sysroots/x86_64-pokysdk-linux/usr/bin/cargo contained in package cargo-cross-canadian-aarch64 requires libcurl.so.4()(64bit), but no providers found in RDEPENDS_cargo-cross-canadian-aarch64? [file-rdeps]`

Adding `nativesdk-libcurl` to `DEPENDS` also solved this (however I would've expected the build to fail because of the missing dependency so it might make sense to double check which headers the build is actually using).

lorenzph:
I don't know to which changes to pin those notes but other things I've observed (after I managed to compile and install the packages to the SDK):

The dynamic loader is wrongly set on all binaries produced by the build (similar to the issue reported above):
```
patchelf --print-interpreter $(which cargo)
/.../build/tmp/sysroots-uninative/x86_64-linux/lib/ld-linux-x86-64.so.2
```
I worked around the manually via patchelf

Builds for the SDK host fail as the linker cannot be found (this can probably be fixed by providing a custom target config for the SDK host referencing the linker provided by nativesdk-gcc):
```
error: linker `cc` not found
  |
  = note: No such file or directory (os error 2)

error: aborting due to previous error

error: could not compile `libc`
```
My reproducer for these issues is an empty project depending on the `bindgen` create (which in turn compiles binaries for the host system)

weltling:
For the first one:

```
$ patchelf --print-interpreter $(which cargo)
/tmp/rust-sdk-deploy-19/sysroots/x86_64-pokysdk-linux/lib/ld-linux-x86-64.so.2
```

How do you run this? Did you install the SDK? Seems there's something very different we're doing.


With the bindgen - there is none in the SDK, or you installed it there? For this implementation, the focus was on `rustc` and `cargo` producing binaries for the target. The nativesdk in general is about producing binaries for the target, not for the host. The currently tested things are here https://github.com/meta-rust/meta-rust/pull/313#issuecomment-782753188

But also, i went for getting bindgen there:

```
$ cargo install bindgen
    Updating crates.io index
  Downloaded bindgen v0.57.0
  Downloaded 1 crate (199.1 KB) in 0.26s
  Installing bindgen v0.57.0
  Downloaded aho-corasick v0.7.15
  Downloaded ansi_term v0.11.0
  Downloaded env_logger v
.......
   Compiling regex v1.4.3
   Compiling cexpr v0.4.0
   Compiling env_logger v0.8.3
    Finished release [optimized] target(s) in 50.49s
  Installing /tmp/rust-sdk-deploy-19/sysroots/core2-64-poky-linux/home/cargo/bin/bindgen
   Installed package `bindgen v0.57.0` (executable `bindgen`)
warning: be sure to add `/tmp/rust-sdk-deploy-19/sysroots/core2-64-poky-linux/home/cargo/bin` to your PATH to be able to run the installed binaries

```

But yeah, that one gets a wrong loader 

```
$ patchelf --print-interpreter /tmp/rust-sdk-deploy-19/sysroots/core2-64-poky-linux/home/cargo/bin/bindgen
/lib/ld-linux-x86-64.so.2
```

Which is actually expected, as that's the loader as in the standard Poky image. For that to work, it is most likely about building `bindgen` for the SDK from a specific recipe. I can't recall the SDK was producing additional tooling, it's  a frozen set of tools. The additional tooling should be on the plan once this one is done. 

Thanks.

weltling:
Looking more at the `bindgen` part - you probably could try to pass the `--target` when building it. But the point is - you still build for the SDK target, not for the host. Say if the target is `aarch64`, it'll build the `bindgen` for `aarch64`. That probably shouldn't work for now in this form. And using bindgen from outside the SDK will have lots of other issues, most likely. The only way i'd see is produce bindgen for the SDK from within the Yocto build.

Thanks.

lorenzph:
> How do you run this? Did you install the SDK? Seems there's something very different we're doing.

I am using a setup where opkg is actually used to install packages directly into the SDK. This works fine as the SDK is installed to the location used during the build process (e.g. /opt/sdk/sysroots/x86_64-pokysdk-linux). However, the binaries produced by the recipes do not set the correct dynamic loader and as such can only be used after invoking the SDK installer (which I believe patches the dynamic loader of all binaries to the installation prefix of the SDK).

> With the bindgen - there is none in the SDK, or you installed it there? For this implementation, the focus was on rustc and cargo producing binaries for the target. The nativesdk in general is about producing binaries for the target, not for the host. The currently tested things are here #313 (comment)

In my use case a project depends on `bindgen` in its Cargo.toml (see https://rust-lang.github.io/rust-bindgen/tutorial-1.html) and this triggers the compilation issue due to the missing target definition for the SDK. I think its fine to leave this out of scope for the initial integration but I also don't see a way to ignore compilation for the SDK target in the long run (or is there another approach which can be used).

weltling:
> I am using a setup where opkg is actually used to install packages directly into the SDK.

Is this documented somewhere? I don't seem to recall such a way to get things done. When the SDK is installed from a shar, the binaries are modified to work under that exact installation path. If something is added, the binaries need to be rebased same way, i'd guess.

> In my use case a project depends on bindgen in its Cargo.toml 

Hmm, tihs seems something different from just running `cargo install ...` in the SDK. Currently it is possible to build for host embedded in the build process and I explicitly worked on that in the regard to build scripts. Fe if a project has `build.rs` like quite some cargo packages do, that file is compiled for host and is executed to probe features and add some options to the build. As an example it's https://github.com/rust-lang/libc/blob/master/build.rs.  This scenario work as a helper program that is built and run on host within the `cargo build` procedure. That is also the reason why the host `libstd` is needed.

There might be something different in how bindgen class works there, need to check on that one. Indeed we're exploring what are the limitations of the Rust usage in the SDK.

Thanks

lorenzph:
> Is this documented somewhere? I don't seem to recall such a way to get things done. When the SDK is installed from a shar, the binaries are modified to work under that exact installation path. If something is added, the binaries need to be rebased same way, i'd guess.

I don't believe that package installation is supported in Yocto SDK builds (although I haven't checked in a while). Nevertheless I wouldn't expect packages to leak paths specific to the build environment (and I am also not aware of any other packages doing so).

weltling:
Yep, what you've shown previously with the output of `patchelf --print-interpreter $(which cargo)` seems that that binary was not patched by the installation shar. From what i could tell, that would happen automatically for all the binaries in the host sysroot (see scripts/relocate_sdk.py within poky).

Thanks.

weltling:
@jmesmon @ColinFinck what is the way to bring this forward, could you advise please? Rebasing 3rd time here :)

Thanks!

jmesmon (approved):
I think we should merge this because it definitely gets us closer to something that's functional, and avoids needing the ongoing rebase.

weltling:
Thanks a lot for checking. I was into some other work last weeks and looks like one more time it'll need to be rebased. However, this topic is definitely on my plan for April, i'm gonna go for another rebase next days and will ping back. It would be great to make  progress and thus have a foundation to stabilize further.

Thanks

jmesmon:
@weltling I've invited you to the @meta-rust/maintainers github org group. If you accept the invite, you'll be able to merge this yourself since another member (me) has approved it.

weltling:
@jmesmon accepted with thanks! Looks like the mail was filtered on my side, thus good you've mentioned it.

I've just rebased and synced with the recent addition of 1.51.0. Probably would be good to repeat some checks with the SDK, i'm intended to do that and proceed with a merge next couple of days.

Thanks

weltling:
I tested the latest state of this patch by producing an SDK for 1.51.0 and repeating the steps like in https://github.com/meta-rust/meta-rust/pull/313#issuecomment-782753188 and some similar manual testing. So far no regression from the previous state with 1.49.0 could be seen. If there's no objection, I'll merge this PR at some point tomorrow.

Thanks

kraj:
> I tested the latest state of this patch by producing an SDK for 1.51.0 and repeating the steps like in [#313 (comment)](https://github.com/meta-rust/meta-rust/pull/313#issuecomment-782753188) and some similar manual testing. So far no regression from the previous state with 1.49.0 could be seen. If there's no objection, I'll merge this PR at some point tomorrow.
> 
> Thanks

Looks fine to me. Please go ahead.

weltling:
Thanks for the reviews. Thanks a good start.

PR #317, opened 2021-02-20
rust: Enable ppc64le target

kraj:
this helps building the compiler, work is still needed for libstd-rs and
other pieces

### DIFF
--- a/recipes-devtools/rust/rust.inc
+++ b/recipes-devtools/rust/rust.inc
@@ -236,6 +236,14 @@ TARGET_POINTER_WIDTH[powerpc] = "32"
 TARGET_C_INT_WIDTH[powerpc] = "32"
 MAX_ATOMIC_WIDTH[powerpc] = "32"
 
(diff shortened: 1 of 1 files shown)


tylerwhall (approved):
The vector alignments aren't included in `--print target-spec-json` for `powerpc64le-unknown-linux-gnu`. Curious when those are required and how they're derived. Seems reasonable and harmless though.

`v256:256:256-v512:512:512`

kraj:
> The vector alignments aren't included in `--print target-spec-json` for `powerpc64le-unknown-linux-gnu`. Curious when those are required and how they're derived. Seems reasonable and harmless though.
> 
> `v256:256:256-v512:512:512`

I have take them from llvm sources 

https://github.com/llvm/llvm-project/blob/main/clang/test/CodeGen/target-data.c#L153-L155

ColinFinck:
Please also check https://github.com/meta-rust/meta-rust/pull/319, where I did a ppc64-specific change.

kraj:
can we merge this

PR #323, opened 2021-02-24
install example executables

jwinarske:
this enables installing example executables without having to override cargo_do_install()


jmesmon (approved):
I was initially curious if we should do this given that examples probably don't want to get installed all the time, but it looks like they'll only get added if someone tweaks their `CARGO_BUILD_FLAGS` to enable building of them.

And if we're building, installing makes sense.

PR #325, opened 2021-03-16
rust.inc: Build the profiler runtime

janderholm:
Needed when compiling with options such as "-C profile-generate" or
"-Z instrument-coverage", e.g. when generating coverage reports or doing
profile guided optimization.


kraj (approved):
lgtm

PR #329, opened 2021-03-28
Enable a workflow to build rust-hello-world

srwalter:
Verify that rust-hello-world build (which in turn ensures that
rust-native, cargo-native, etc all build)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,23 @@
+on: [pull_request]
+jobs:
+  build:
+    runs-on: self-hosted
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: 'meta-rust'
+      - name: Fetch poky
+        run: |
+          mv poky/build/sstate-cache . || true
+          rm -rf poky meta-openembedded
+          git clone -b gatesgarth --single-branch git://git.yoctoproject.org/poky
+          git clone -b gatesgarth --single-branch git://git.openembedded.org/meta-openembedded
+      - name: Run bitbake
+        run: |
+          cd poky
+          . oe-init-build-env
+          mv ../../sstate-cache . || true
+          sed -i -e "\$i$(readlink -f ../../meta-openembedded/meta-oe) \\\\" conf/bblayers.conf
+          sed -i -e "\$i$(readlink -f ../../meta-rust) \\\\" conf/bblayers.conf
+          echo 'PARALLEL_MAKE_pn-rust-llvm-native = "-j2"' >> conf/local.conf
+          bitbake rust-hello-world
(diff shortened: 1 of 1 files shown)


jmesmon (approved):
Nice. going this way (self hosted gh runners) seems reasonable.

Do you have details on the vm setup so others can add their own runners? I was working on some kickstarter file setup for a similar setup but didn't finish it.

.github/workflows/main.yml
```diff
@@ -0,0 +1,23 @@
...
+          rm -rf poky meta-openembedded
+          git clone -b gatesgarth --single-branch git://git.yoctoproject.org/poky
+          git clone -b gatesgarth --single-branch git://git.openembedded.org/meta-openembedded
+      - name: Run bitbake
+        run: |
+          cd poky
+          . oe-init-build-env
+          mv ../../sstate-cache . || true
+          sed -i -e "\$i$(readlink -f ../../meta-openembedded/meta-oe) \\\\" conf/bblayers.conf
+          sed -i -e "\$i$(readlink -f ../../meta-rust) \\\\" conf/bblayers.conf
```
jmesmon, line 21: I think there's a `bitbake-layers` tool that we can use for this.

srwalter:
> Nice. going this way (self hosted gh runners) seems reasonable.
> 
> Do you have details on the vm setup so others can add their own runners? I was working on some kickstarter file setup for a similar setup but didn't finish it.

This VM is just a basic Ubuntu Server 20.10 install with only the packages required by bitbake installed (build-essential, chrpath, and diffstat I believe).  Then I followed the directions in the Github docs for running your own self-hosted runner

jmesmon:
Oh, related: because these builds take so long, and to ensure there's some level of review before we run arbitrary code on folks gh-runners, we could use bors-ng to trigger running the builds. Not required for this, just something to think about for the future.

srwalter:
> Oh, related: because these builds take so long, and to ensure there's some level of review before we run arbitrary code on folks gh-runners, we could use bors-ng to trigger running the builds. Not required for this, just something to think about for the future.

Yes that's probably a good thing to look into.  This is a new VM with nothing else on it and firewalled off from the rest of the world, so it would only be an annoyance if someone did try to hijack it.  But still, better to not be annoyed at all :)

.github/workflows/main.yml
kraj: I use GH actions for CI for YoeDistro regularly and also for other layers I maintain see

see how its done for meta-clang, where it checks out distro layers and replaces meta-clang with the pull branch smartly.
https://github.com/kraj/meta-clang/blob/master/.github/workflows/yoe.yml

yoe CI
https://github.com/YoeDistro/yoe-distro/blob/master/.github/workflows/yoe.yml

maybe it will be helpful

PR #333, opened 2021-04-02
workflows: Build and test SDK

### LINKED ISSUE #332: Exercise SDK in CI
Now that we have SDK support, we should verify that it continues to work in our CI.

weltling:
This adds a basic SDK functionality test including build, installation and a test crate build.

Closes #332.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,7 @@ jobs:
           bitbake rust-hello-world
       - name: Test SDK
         run: |
+          SDK_DIR=$PWD/sdk-test
           cd poky
           . oe-init-build-env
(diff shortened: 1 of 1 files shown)


jmesmon (approved):
May want to use a directory under the GitHub runner workspace instead of in /tmp for the sdk install? Thinking about if we'll have leftovers or space consumption issues

weltling:
Thanks for the hint. I've also reduced the size a bit more. As the run passes, it might be acceptable, whereby I'm not very happy with the >1h test runtime. More things might come. I'll check to perhaps reduce the size more, maybe also use the official sstate mirror, disable recommendations, etc. Either in the scope of this one, or as a follow up.

Thanks

PR #335, opened 2021-04-04
workflows: Refactor yml to improve build time

weltling:
- Split out the build configuration job
- Use the official sstate mirror
<strike>- Disable more of unnecessary features</strike>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,9 @@
 on: [pull_request]
 jobs:
   build:
+    env:
+      YOCTO_VERSION: 3.2.3
+      YOCTO_BRANCH: gatesgarth
     runs-on: self-hosted
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


weltling:
Hmm, looks like this is the fastest run and not reducing the build doesn't cause storage overflow issues. I'd guess the official `sstate` is picked up, so this approach might indeed serve. Switching to fixed poky release tags will be needed, however.

Thanks.

srwalter (approved):

srwalter:
@jmesmon I noticed "squash and merge" is disabled in this project.  I'm not sure who changed it, do you have an opinion on its use?

jmesmon:
I'm unopposed to enabling/allowing it 🤷‍♂️

It's nice sometimes to have that option so folks don't need to locally squash.

weltling:
Thanks for the checks!

A case having to push incremental changes to satisfy CI is always about piling an onion. Requiring to squash the original change set would be certainly justified in such a case.

Generalizing might be good enough, too. Requiring to merge just one commit (and description) could be more conceptive. Still it's up to whether on would like to keep the history.

Thanks

PR #338, opened 2021-04-21
rust.inc: add strict-align when target is ARMv4

### LINKED ISSUE #337: Alignement issues on ARMv4
## Version(s) of meta-rust
master (a4302679bbe2bfc5762aa47003ca8192f21a1342)

master/hardknott (7d0988966cc1fd9e93549ca2ec21ac3bc765ceb6)

## Expected result
Complied code runs fine on target hardware

## Actual result
Code crashes with a "Bus Error"

## Steps to reproduce

Compile a non trivial program an try to run it on an ARMv4 cpu

aj-bagwell:
Not sure if this is the correct way to do this, but this fixes #337

### DIFF
--- a/recipes-devtools/rust/rust-common.inc
+++ b/recipes-devtools/rust/rust-common.inc
@@ -39,6 +39,9 @@ def llvm_features_from_tune(d):
     if ('armv6' in mach_overrides) or ('armv6' in feat):
         f.append("+v6")
 
+    if ('armv4' in mach_overrides) or ('armv4' in feat):
+        f.append("+strict-align")
+
     if 'dsp' in feat:
         f.append("+dsp")
 


ColinFinck:
Looking at https://github.com/rust-lang/rust/tree/master/compiler/rustc_target/src/spec, Rust adds `+strict-align` for almost all ARM-based targets, not just ARMv4.
I suggest adding it globally to `FEATURES[arm]`, `FEATURES[armv7]`, and `FEATURES[aarch64]`, not conditionally.

Any other reviewer opinion?

kraj:
> Looking at https://github.com/rust-lang/rust/tree/master/compiler/rustc_target/src/spec, Rust adds `+strict-align` for almost all ARM-based targets, not just ARMv4.
> I suggest adding it globally to `FEATURES[arm]`, `FEATURES[armv7]`, and `FEATURES[aarch64]`, not conditionally.
> 
> Any other reviewer opinion?

newer ARM ( v6+) have hardware support for handling unaligned access, So it does not matter as much but I guess using strict align will be better perhaps code size might increase a little. I think lets merge this change and then do the rest as a follow up.

srwalter (approved):

PR #340, opened 2021-04-21
Don't include target libs in linker path for build.rs

### LINKED ISSUE #339: Crate depending on `zmq` fail to compile when targeting musl
## Version(s) of meta-rust
master (a4302679bbe2bfc5762aa47003ca8192f21a1342)

master/hardknott (7d0988966cc1fd9e93549ca2ec21ac3bc765ceb6)

## Expected result
package compile fine

## Actual result
Complile fails with a link error

## Steps to reproduce

Compile a package that has a non trivial C dependency for example something that depends on [zmq](https://crates.io/crates/zmq) with `TCLIBC = "musl"`

aj-bagwell:
When compiling a crate that relies on c libaries for a musl target
cargo will use the same link paths for both the build.rs and the
crate, this means that the build.rs which needs to be run on the host
gets linked to libraries that are compiled for the target which can cause
issues.

This hack filters out non '-native' sysroots from the link path for build.rs

fixes #339

### DIFF
--- a/classes/rust-common.bbclass
+++ b/classes/rust-common.bbclass
@@ -138,6 +138,26 @@ create_wrapper () {
 	chmod +x "${file}"
 }
 
+create_build_wrapper () {
+	file="$1"
+	shift
+
+	cat <<- EOF > "${file}"
+	#!/usr/bin/env bash
+
+	args=("\$@")
+	for ((i=0; i<"\${#args[@]}"; ++i)); do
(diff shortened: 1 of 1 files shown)


jmesmon (changes requested):
so we definitely don't want to do this. but we do want to fix the problem of nested build systems (or at least decide what the approach will be), which I think #339 should let us discuss.

PR #345, opened 2021-05-06
nativesdk: Add some tools in a separate recipe

weltling:
This brings ATM:
- clippy
- rustfmt

Another approach might be to include these into the existing rust recipe,
but as we can't really use PACKAGECONFIG here it might just overload it.
This way, the tools will be built in parallel and still can be excluded
from the packgae group, if desired.

### DIFF
--- a/recipes-devtools/rust/rust-cross-canadian-common.inc
+++ b/recipes-devtools/rust/rust-cross-canadian-common.inc
@@ -1,0 +1,159 @@
+
+RUST_ALTERNATE_EXE_PATH = "${STAGING_LIBDIR_NATIVE}/llvm-rust/bin/llvm-config"
+
+require rust-target.inc
+
+inherit cross-canadian
+
+DEPENDS += "  \
+            virtual/${HOST_PREFIX}gcc-crosssdk \
+            virtual/nativesdk-libc rust-llvm-native \
+            virtual/${TARGET_PREFIX}compilerlibs \
+            virtual/nativesdk-${HOST_PREFIX}compilerlibs \
+            gcc-cross-${TARGET_ARCH} \
+           "
+
+# The host tools are likely not to be able to do the necessary operation on
+# the target architecturea. Alternatively one could check compatibility
+# between host/target.
+EXCLUDE_FROM_SHLIBS_${RUSTLIB_TARGET_PN} = "1"
+
+DEBUG_PREFIX_MAP = "-fdebug-prefix-map=${WORKDIR}=/usr/src/debug/${PN}/${EXTENDPE}${PV}-${PR} \
+                    -fdebug-prefix-map=${STAGING_DIR_HOST}= \
+                    -fdebug-prefix-map=${STAGING_DIR_NATIVE}= \
+                    "
+
+LLVM_TARGET[x86_64] = "${RUST_HOST_SYS}"
+python do_rust_gen_targets () {
+    wd = d.getVar('WORKDIR') + '/targets/'
+    rust_gen_target(d, 'TARGET', wd, d.getVar('TARGET_LLVM_FEATURES') or "", d.getVar('TARGET_LLVM_CPU'), d.getVar('TARGET_ARCH'))
+    rust_gen_target(d, 'HOST', wd, "", "generic", d.getVar('HOST_ARCH'))
+    rust_gen_target(d, 'BUILD', wd, "", "generic", d.getVar('BUILD_ARCH'))
+}
+
+INHIBIT_DEFAULT_RUST_DEPS = "1"
+
+export TARGET_CC = "${CCACHE}${TARGET_PREFIX}gcc --sysroot=${STAGING_DIR_TARGET} ${TARGET_CC_ARCH} ${SECURITY_NOPIE_CFLAGS}"
+export TARGET_CXX = "${CCACHE}${TARGET_PREFIX}g++ --sysroot=${STAGING_DIR_TARGET} ${TARGET_CC_ARCH} ${SECURITY_NOPIE_CFLAGS}"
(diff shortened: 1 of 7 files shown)


recipes-devtools/rust/rust-cross-canadian-common.inc
```diff
@@ -0,0 +1,159 @@
...
+}
+
+INHIBIT_DEFAULT_RUST_DEPS = "1"
+
+export TARGET_CC = "${CCACHE}${TARGET_PREFIX}gcc --sysroot=${STAGING_DIR_TARGET} ${TARGET_CC_ARCH} ${SECURITY_NOPIE_CFLAGS}"
+export TARGET_CXX = "${CCACHE}${TARGET_PREFIX}g++ --sysroot=${STAGING_DIR_TARGET} ${TARGET_CC_ARCH} ${SECURITY_NOPIE_CFLAGS}"
+export TARGET_CCLD = "${TARGET_PREFIX}gcc --sysroot=${STAGING_DIR_TARGET} ${TARGET_CC_ARCH} ${SECURITY_NOPIE_CFLAGS}"
+export TARGET_AR = "${TARGET_PREFIX}ar"
+
+do_rust_create_wrappers () {
```
srwalter, line 41: This is a lot of duplication of the version in rust-common.bbclass.  Can you refactor things so there's less duplication?

recipes-devtools/rust/rust-cross-canadian-common.inc
```diff
@@ -0,0 +1,159 @@
...
+	create_wrapper "${RUST_TARGET_CC}" "${TARGET_CC}"
+	# Yocto Target / Rust Target C++ compiler
+	create_wrapper "${RUST_TARGET_CXX}" "${TARGET_CXX}"
+	# Yocto Target / Rust Target linker
+	create_wrapper "${RUST_TARGET_CCLD}" "${TARGET_CCLD}" "${TARGET_LDFLAGS}"
+	# Yocto Target / Rust Target archiver
+	create_wrapper "${RUST_TARGET_AR}" "${TARGET_AR}"
+}
+
+python do_configure() {
```
srwalter, line 63: Similarly, we need to find a way for this to be common with the code in rust.inc used to build the compiler.

recipes-devtools/rust/rust-cross-canadian-common.inc
```diff
@@ -0,0 +1,159 @@
...
+
+    # set up ${WORKDIR}/cargo_home
+    bb.build.exec_func("setup_cargo_environment", d)
+}
+
+INSANE_SKIP_${RUSTLIB_TARGET_PN} = "file-rdeps arch ldflags"
+SKIP_FILEDEPS_${RUSTLIB_TARGET_PN} = "1"
+
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+INHIBIT_PACKAGE_STRIP = "1"
```
srwalter, line 157: Why inhibit strip?

srwalter (changes requested):

recipes-devtools/rust/rust-cross-canadian-common.inc
weltling: Thanks for the reviews. There's currently no refactoring in this patch, just pulling out the common parts so they can be reused across recipes. I'll check to do some refactoring as suggested, otherwise i think a wider refactoring might be good for a separate pull request.

Thanks

otavio: I'd suggest you do the rework and send it here as a baseline commit so we have it split for easier review but part of this very single PR.

weltling: Yeah, a small refactoring should be acceptable in the scope of this patch.

Thanks.

recipes-devtools/rust/rust-cross-canadian-common.inc
weltling: Fixed, thanks.

recipes-devtools/rust/rust-cross-canadian-common.inc
weltling: Fixed, thanks.

recipes-devtools/rust/rust-cross-canadian-common.inc
weltling: Regarding inhibiting strip, there are at least these points:

- As the SDK build will produce libstd for both host and target, but the strip tool is available only for host, it won't recognize the binary format for target. Perhaps there's a solution for this, but i'm not sure binutils for target wolud work even if it'd be possible to build them. This is anyway not relevant for any linking at the SDK usage time.
- The toolchain itself doesn't need a separate debug info.

The same approach is used in the golang recipes, as i'd guess there's a similar set of issues where just not stripping is the simplest solution.

Thanks

weltling:
@srwalter @otavio the current state should be fine for the next round.

Thanks

recipes-devtools/rust/rust.inc
```diff
@@ -117,11 +117,13 @@ python do_configure() {
...
-    targets = [d.getVar("TARGET_SYS", True)]
+    if not "targets" in locals():
+        targets = [d.getVar("TARGET_SYS", True)]
     config.set("build", "target", e(targets))
 
-    hosts = [d.getVar("HOST_SYS", True)]
-    config.set("build", "host", e(targets))
+    if not "hosts" in locals():
+        hosts = [d.getVar("HOST_SYS", True)]
+    config.set("build", "host", e(hosts))
```
srwalter, line 126: Looks like this was an existing copy/paste bug.  Nice!

srwalter (approved):

recipes-devtools/rust/rust.inc
weltling: Yeah, it was revealed during the factorization, that's how the code duplication elimination pays off itself :)

Thanks!

PR #353, opened 2021-07-02
rust-common.inc: Use == operator instead of 'is'

kraj:
literal is being checked here therefore == and != should be used instead
of 'is' or 'is not'

### DIFF
--- a/recipes-devtools/rust/rust-common.inc
+++ b/recipes-devtools/rust/rust-common.inc
@@ -52,7 +52,7 @@ def llvm_features_from_tune(d):
         f.append("+dsp")
 
     if 'thumb' in feat:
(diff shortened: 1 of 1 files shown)


weltling (approved):

akuster:
I am seeing this issue too.

weltling:
I'd suggest we should merge this as the CI seems still broken :/ I'll check to do so tomorrow, if no one beats me to it.

The log tells 

"This request was automatically failed because there were no enabled runners online to process the request for more than 1 days."

but no details. Perhaps I miss something due to to permission, but otherwise there's not more than that. Perhaps @srwalter could you check on that, too?

Thanks

PR #359, opened 2021-08-08
Upgrade to 1.54.0

avrabe:
Upgrade to 1.54.0
    
The rust patches for riscv64 have been upstreamed
and are not needed anymore.

The riscv64 patches for libstd-rs need to be
applied to libc-0.2.93 to work. Some slight
modifications are necessary.

Signed-off-by: Ralf Anton Beier <ralf_beier@me.com>


srwalter:
Rather than renaming 1.51 to 1.54, I think the way we've been operating would be to keep 1.51 and drop 1.49.  Do any other maintainers want to chime in if I'm remembering the policy correctly?

ColinFinck:
> Rather than renaming 1.51 to 1.54, I think the way we've been operating would be to keep 1.51 and drop 1.49. Do any other maintainers want to chime in if I'm remembering the policy correctly?

Exactly, 1.51 and 1.54 should remain.

weltling:
Same here, things move rapidly and 1.49 might be of less use today and later Rust should be compatible when same edition is used. Maintained projects would however profit more from possibly stabilized features in a later version.

Thanks

avrabe:
Changed it to drop 1.49.0 and to keep 1.51.0. 
Compiled both versions mainly with the `MACHINE=qemuriscv64` and `TCLIB=musl` to verify that the patches still apply.

kraj (approved):
lgtm

shr-project:
I've sent similar cleanup to oe-core https://lists.openembedded.org/g/openembedded-core/message/155617

shr-project:
https://github.com/meta-rust/meta-rust/pull/363 will remove the unused patches

PR #363, opened 2021-09-03
Remove unused patch file

avrabe:
Rust 1.49.0 needed the 0001-rustc_target-Fix-dash-vs-underscore-mismatches-in-op.patch
patch and newer versions not.

Signed-off-by: Ralf Anton Beier <ralf_beier@me.com>


shr-project:
recipes-devtools/rust/libstd-rs-1.51.0/0004-Update-checksums-for-modified-files.patch is also unused (was used only be 1.49.0 version and then incorrectly moved to 1.51.0 directory when 1.49.0 was removed).

srwalter (approved):

PR #364, opened 2021-09-03
meta-rust: clean-up now that much of meta-rust is in oe-core

rwmacleod:
This builds when using meta-rust and poky/master  and having a conf/local.conf file with:
require conf/distro/include/rust_versions.inc

BUT, there's a warning about cargo-native:
WARNING: preferred version 1.51.0 of cargo-native not available (for item cargo-native)
WARNING: versions of cargo-native available: 1.54.0

which is  odd. Looking at the cooker logs all other rust toolchain components are using version 1.51
and building rust-hello-world works fine.
I'm debugging but I wanted to make this clean-up available in case anyone wanted it.


rwmacleod:
I'm not able to reproduce the selection of cargo-native-1.54 on another machine for a variety of builds and debugging the version selection in bitbake/lib/bb/providers.py is going to take some time. It would be great if someone else could try to reproduce what I'm seeing.

rwmacleod:
FYI, with a sligh change, I don't see this now.

I suggested removing rust_versions.inc from oe-core but @kraj  asked that I move it's contents to meta/conf/distro/include/tcmode-default.inc. I've done that an with a local build and RUSTVERSION = "1.51.0" in my local.conf, I'm no longer picking up cargo-1.54.0. I don't think it will hurt anything to have the extra rust_versions.inc file in meta-rust so I'll send a PR to remove it if this gets merged.

rwmacleod:
How do people feel about this PR? Do we want or need this meta-rust layer still ?

NathanNZ:
Wouldn't this remove functionality for dunfell? I believe it's unlikely that the oe-core will be backported - I feel like it would still be a useful package to have those who are sticking with LTS versions accordingly?

otavio:
I'd branch to dunfell and move master on, so people wishing to use with older release can use with the branch.

ColinFinck:
I just talked to @rossburton on the #oe IRC channel and he said
> ask the meta-rust maintainers if you want that to have a backport of new rust
> part of the expectation of the LTS releases is that people with eg an interest in latest rust for dunfell will maintain a layer which simply contains latest rust for dunfell

Thinking this through, we won't just need meta-rust for dunfell, but also for every other Yocto release where oe-core doesn't provide a Rust version or has no interest in backporting the latest Rust version.

snke:
Would it also make sense to make somekind of note in the README that this layer is integrated into oe and you don't need it when using honister or above?

rzr:
Or add explicity branches supported in layer.conf

PR #382, opened 2022-05-10
Add support for Cargo features and flags

sirhcel:
- I was missing this feature for development builds to enable debug features of an application
- Adds support for enabling disabling Cargo features when building the application through new variables
- Adds support for passing generic flags/arguments to Cargo via `CARGO_EXTRA_FLAGS`
- The newly introduced variables aim to achieve some compatibility with [meta-rust-bin](https://github.com/rust-embedded/meta-rust-bin)
    - Using `CARGO_FEATURES` for feature selection
    - Providing `CARGO_EXTRA_FLAGS` for an alternative way setting `--all-features` and `--no-default-features`

### DIFF
--- a/classes/cargo.bbclass
+++ b/classes/cargo.bbclass
@@ -29,9 +29,29 @@ CARGO_SRC_DIR ??= ""
 # The actual path to the Cargo.toml
 MANIFEST_PATH ??= "${S}/${CARGO_SRC_DIR}/Cargo.toml"
 
+# Features and additional flags for 'cargo build'.
+#
+# CARGO_FEATURES supports both, a comma or space separated list. Disabling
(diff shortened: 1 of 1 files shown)


sirhcel:
What about backporting this change to older release branches as well?

cletustboone:
Also need this, so hopefully it goes in soon.

sirhcel:
@cardoe, do you have any objections against adding feature flags or suggestions for improving this PR?

dougcooper:
would very much like to see this merged given when it was opened. it could be simplified to just the inclusion of `${EXTRA_CARGO_FLAGS}` if there is a concern.

srwalter (approved):

PR #390, opened 2022-09-08
rust-common.inc: +v7 implies +neon

srwalter:
If we don't support neon, then (at least on armv7) we must explicitly throw -neon.  Additionally, this must occur after the "+v7" to have an effect.


otavio:
Have you send same fix for OE-Core?

otavio:
Please send it over ML.

PR #391, opened 2022-09-09
Fix github CI

srwalter:


tylerwhall (approved):
I suppose it's better to have working CI if the test couldn't be easily fixed.

PR #402, opened 2023-01-05
Add langdale support

g0hl1n:
Add langdale to LAYERSERIES_COMPAT of rust-layer.


g0hl1n:
This doesn't work for all architectures... Therefore close it for now.

DavidAntliff:
Out of curiosity, which architectures does it not work with?

Did this PR work with `cortexa72-cortexa53-xilinx-linux`? If not, is there a known workaround to get the main branch working with this architecture in Langdale?

PR #433, opened 2023-10-26
Add Rust 1.73.0

be-neth:
- Bump LLVM_RELEASE to 17.0.2 (seen in github rust project `llvm-project/` submodule).
- Update default version from 1.72.0 to 1.73.0


be-neth:
It fail for dunfell as the required cmake version is not reached for rust-llvm-native (need >=3.20.0 but dunfell provide 3.16.5)

PR #441, opened 2024-02-04
Fix incorrect target json for armv8 / aarch64-poky-linux

tronical:
As seen in https://github.com/briansmith/ring/issues/1873, targeting aarch64-poky-linux may not always explicitly have neon in TUNE_FEATURES, but it should also not be removed because it's... there.

With OECore's rust build that works because it's only marked as absent for armv7:

https://github.com/openembedded/openembedded-core/blob/97b5d0c85af0b667854eea90ace0a8c2f51ecefe/meta/classes-recipe/rust-target-config.bbclass#L32

So apply the same here.


tronical:
Thank you for the quick review and merge ❤️

PR #448, opened 2024-06-27
cargo_common: Exclude http_proxy from variable dependencies

lorenzph:
The value of `http_proxy` does not affect the build output and as such should be excluded from the task's hash to avoid sstate cache misses on build hosts with different `http_proxy` settings.

### DIFF
--- a/classes/cargo_common.bbclass
+++ b/classes/cargo_common.bbclass
@@ -105,6 +105,7 @@ cargo_common_do_configure () {
 	progress.width = 80
 	EOF
 }
+cargo_common_do_configure[vardepsexclude] += "http_proxy"
 
 oe_cargo_fix_env () {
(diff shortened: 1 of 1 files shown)


lorenzph:
I'm moving this back into draft state as `http_proxy` should really be ignored via `BB_HASHCONFIG_IGNORE_VARS`. I'll try to figure out why this doesn't seem to work first.

lorenzph:
It turns out I confused `BB_HASHCONFIG_IGNORE_VARS` which is relevant to the recipe metadata cache and `BB_BASEHASH_IGNORE_VARS` which is relevant to the task hash. The latter doesn't include `http_proxy` which explains the observed behaviour.

srwalter (approved):

Issue #449, opened 2024-07-30
Kirkstone build fails to build version 1.78


maskimthedog:
## Version(s) of meta-rust
1.78

## Version(s) of poky and/or oe-core
Kirkstone

## Expected result
Clean build

## Actual result
```
ve_gdk_pixbuf --cfg system_deps_have_gio --cfg system_deps_have_glib --cfg system_deps_have_harfbuzz --cfg system_deps_have_libxml2 --cfg system_deps_have_pangocairo --cfg system_deps_have_pangoft2`
| warning: method `fixed_eq_cairo` is never used
|    --> src/float_eq_cairo.rs:32:8
|     |
| 31  | pub trait FixedEqCairo {
|     |           ------------ method in this trait
| 32  |     fn fixed_eq_cairo(&self, other: &Self) -> bool;
|     |        ^^^^^^^^^^^^^^
|     |
| note: the lint level is defined here
|    --> src/lib.rs:145:46
|     |
| 145 | #![warn(nonstandard_style, rust_2018_idioms, unused)]
|     |                                              ^^^^^^
|     = note: `#[warn(dead_code)]` implied by `#[warn(unused)]`
|
| error: casting references to a bigger memory layout than the backing allocation is undefined behavior, even if the reference is unused
|    --> src/surface_utils/mod.rs:154:13
|     |
| 153 |             let p: *mut u8 = &mut self[y as usize * stride + x as usize * 4];
|     |                                   ------------------------------------------ backing allocation comes from here
| 154 |             *(p as *mut u32) = value;
|     |             ^^^^^^^^^^^^^^^^^^^^^^^^
|     |
|     = note: casting from `u8` (1 bytes) to `u32` (4 bytes)
|     = note: `#[deny(invalid_reference_casting)]` on by default
|
| warning: unused return value of `Box::<T>::from_raw` that must be used
|    --> src/xml/xml2_load.rs:371:5
|     |
| 371 |     Box::from_raw(ctx);
|     |     ^^^^^^^^^^^^^^^^^^
|     |
|     = note: call `drop(Box::from_raw(ptr))` if you intend to drop the `Box`
|     = note: `#[warn(unused_must_use)]` implied by `#[warn(unused)]`
| help: use `let _ = ...` to ignore the resulting value
|     |
| 371 |     let _ = Box::from_raw(ctx);
|     |     +++++++
|
| warning: `librsvg` (lib) generated 5 warnings (3 duplicates)
[...]

PR #452, opened 2024-10-15
Add crosssdk support for kirkstone build

raennor:
### Version(s) of meta-rust
latest master

### Version(s) of poky and/or oe-core
kirkstone

### Expected result
Building of toolchain which includes nativesdk-python3-cryptography  should be finished without error

### Actual result
Build failed with following error:
```
| warning: build_rust: Forced Rust target `x86_64-pokysdk-linux` is not compatible with deduced Rust target `x86_64-unknown-linux-gnu` - the built package  may not import successfully once installed.
|
| error: Error loading target specification: Could not find specification for target "x86_64-pokysdk-linux". Run `rustc --print target-list` for a list of built-in targets
```

### Fix for this
To fix this issue I added the rust-crosssdk for all rust versions which are supported on this layer.

I also notice that librsvg 2.52.10 building failed (which is also referred on https://github.com/meta-rust/meta-rust/issues/449), which should be fixed with my second commit.


recipes-gnome/librsvg/librsvg/0002-allow-invalid-ref-casting-to-allow-building.patch
```diff
@@ -0,0 +1,20 @@
+--- librsvg-2.52.10/src/surface_utils/mod.rs	2024-10-15 20:59:07.374334093 +0200
++++ librsvg-2.52.10/src/surface_utils/mod.rs_mod	2024-10-15 20:58:56.391383870 +0200
+@@ -147,12 +147,12 @@
+     #[inline]
+     fn set_pixel(&mut self, stride: usize, pixel: Pixel, x: u32, y: u32) {
+         let value = pixel.to_u32();
++        let index = (y as usize + stride) + (x as usize * 4);
```
srwalter, line 7: Shouldn't this be `y as usize * stride`?

recipes-gnome/librsvg/librsvg/0002-allow-invalid-ref-casting-to-allow-building.patch
```diff
@@ -0,0 +1,20 @@
...
+ 
+-        #[allow(clippy::cast_ptr_alignment)]
+-        unsafe {
+-            let p: *mut u8 = &mut self[y as usize * stride + x as usize * 4];
+-            *(p as *mut u32) = value;
+-        }
++        assert!(index + 4 <= self.len(), "Index out of bounds");
```
srwalter, line 14: This check is both incorrect and unnecessary.  copy_from_slice will do bounds checking.

srwalter (changes requested):

recipes-gnome/librsvg/librsvg/0002-allow-invalid-ref-casting-to-allow-building.patch
raennor: Oh, yes you are absolutely correct, this was a typo.

recipes-gnome/librsvg/librsvg/0002-allow-invalid-ref-casting-to-allow-building.patch
raennor: Thanks for this information. I didn't know that copy_from_slice do bounds checking, so I will remove this check.

srwalter:
Does your patch also apply to the version of librsvg in mickledore?  Consider using librsvg_%.bbappend as the name, if so

raennor:
Hi @srwalter,
sorry, but this patch will only apply until Version 2.53.1 because this function **DerefMut** was removed in version 2.53.2.

raennor:
Hi @srwalter,
I have modified the bbappend so that it will be applied also to mickledore (librsvg_%.bbappend ) but the patch and  and the TARGET_CC_ARCH will only be expanded if PV is 2.52.10 (which is present in kirkstone).

This should possibly fix the workflow issue on mickledore.

srwalter (approved):

PR #453, opened 2024-11-05
Rust 1.82

jaskij:
This necessitated some more changes than a normal update, since `changelog-seen` is for older version. The added support for `change-id` is split into a separate commit. The only info I have been able to find online is from was from the [Chromium issue tracker](https://issues.chromium.org/issues/335782507)

### DIFF
--- a/recipes-devtools/cargo/cargo-cross-canadian_1.82.0.bb
+++ b/recipes-devtools/cargo/cargo-cross-canadian_1.82.0.bb
@@ -1,0 +1,6 @@
+require recipes-devtools/rust/rust-source-${PV}.inc
+require recipes-devtools/rust/rust-snapshot-${PV}.inc
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/cargo-${PV}:"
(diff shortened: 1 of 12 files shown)


jaskij:
@srwalter seems your runner ran out of disk space

srwalter (approved):

brennan-brt:
Shouldn't this commit have also updated? @jaskij @srwalter  https://github.com/meta-rust/meta-rust/blob/5df53282955b0a8dc1c4e28e56091e53b2f8c59e/conf/distro/include/rust_versions.inc#L4

jaskij:
Personally, I don't care either way, since IMO `RUST_VERSION` should be set by your distro @brennan-brt

brt-adam-snaider:
@jaskij I think the defaults as they are right now break `rustc` compilation in unexpected ways. Seems reasonable to remove the default version but it should probably be an explicit error if RUST_VERSION is unset as opposed to failing to build rustc due to some kind of version incompatibility.

PR #455, opened 2024-11-12
rust: 1.82.0: Fix builds for aarch64 targets

b49020:
Rust data layout for aarch64 updated in 1.80.0 release following LLVM data layout update (see: https://github.com/rust-lang/rust/pull/124813).

So conditionally update data layout for aarch64 target build as well.

### DIFF
--- a/recipes-devtools/rust/rust-common.inc
+++ b/recipes-devtools/rust/rust-common.inc
@@ -140,7 +140,7 @@ FEATURES[armv7-eabi] = "+v7,+vfp2,+thumb2"
 
 ## aarch64-unknown-linux-{gnu, musl}
 DATA_LAYOUT[aarch64] = "e-m:e-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128"
-DATA_LAYOUT[aarch64] .= "${@ "" if bb.utils.vercmp_string_op(d.getVar('RUST_VERSION', True), "1.80.0", "<") else "-Fn32"}"
+DATA_LAYOUT[aarch64] .= "${@'' if bb.utils.vercmp_string_op(d.getVar('RUST_VERSION', True), '1.80.0', '<') else '-Fn32'}"
 LLVM_TARGET[aarch64] = "${RUST_TARGET_SYS}"
 TARGET_ENDIAN[aarch64] = "little"
 TARGET_POINTER_WIDTH[aarch64] = "64"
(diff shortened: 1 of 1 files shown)


b49020:
@jaskij @srwalter fyi..

jaskij:
Thanks for catching that. This touches on the deeper parts of the toolchain I'm not familiar with, so no further comment from me.

srwalter:
Looks like it no build.

b49020:
> Looks like it no build.

Strangely it build fine for me locally. I have tried to invert the comparison operation, can you trigger the build again?

albert2004:
Issue is also visible on my side (missing Fn32 in DATA_TYPES). Looking at verification result, it looks like yocto syntax is not processed and instead of adding "-Fn32" whole query is added to string.

be-neth:
Tested on kirkstone, I have failed to build with this PR directly, but another try by replacing double quote by single one in your PR seems to work.

```
DATA_LAYOUT[aarch64] .= "${@'' if bb.utils.vercmp_string_op(d.getVar('RUST_VERSION', True), '1.80.0', '<') else '-Fn32'}"
```

recipes-devtools/rust/rust-common.inc
```diff
@@ -140,6 +140,7 @@ FEATURES[armv7-eabi] = "+v7,+vfp2,+thumb2"
 
 ## aarch64-unknown-linux-{gnu, musl}
 DATA_LAYOUT[aarch64] = "e-m:e-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128"
+DATA_LAYOUT[aarch64] .= "${@ "" if bb.utils.vercmp_string_op(d.getVar('RUST_VERSION', True), "1.80.0", "<") else "-Fn32"}"
```
srwalter, line 143:
```suggestion
DATA_LAYOUT[aarch64] .= "${@'' if bb.utils.vercmp_string_op(d.getVar('RUST_VERSION', True), '1.80.0', '<') else '-Fn32'}"
```

srwalter (changes requested):

be-neth:
There is something strange with RUST_VERSION definition ?
Builder are complaining that there is no version to strip ?

On my side, with a RUST_VERSION defined somewhere in my layer.conf its working.

recipes-devtools/rust/rust-common.inc
```diff
@@ -140,6 +140,7 @@ FEATURES[armv7-eabi] = "+v7,+vfp2,+thumb2"
 
 ## aarch64-unknown-linux-{gnu, musl}
 DATA_LAYOUT[aarch64] = "e-m:e-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128"
+DATA_LAYOUT[aarch64] .= "${@'' if bb.utils.vercmp_string_op(d.getVar('RUST_VERSION', True), '1.80.0', '<') else '-Fn32'}"
```
srwalter, line 143:
```suggestion
DATA_LAYOUT[aarch64] .= "${@'' if bb.utils.vercmp_string_op(d.getVar('PV', True), '1.80.0', '<') else '-Fn32'}"
```

srwalter (approved):

PR #465, opened 2025-05-14
Rust 1.86.0

be-neth:
Update script/rust-get-stage0 for getting snapshot stage0 sha256sum from script.

Add rust 1.86.0

Warning: Update RUST_VERSION from 1.78.0 to 1.86.0.


be-neth:
@srwalter it seems your github workers are not reachable.

I have tested and validated this on Yocto version kirkstone.

